PR #111, opened 2015-01-28
Add logging with null logger

brandondunne:
Depends on https://github.com/ManageIQ/linux_admin/pull/112

### DIFF
--- a/lib/linux_admin/common.rb
+++ b/lib/linux_admin/common.rb
@@ -2,6 +2,12 @@ require 'awesome_spawn'
 
 module LinuxAdmin
   module Common
+    attr_writer :logger
+
+    def logger
+      @logger ||= LinuxAdmin.logger
+    end
+
     def cmd(cmd)
       Distros.local.command(cmd)
     end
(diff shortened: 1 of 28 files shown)


lib/linux_admin/common.rb
```diff
@@ -11,6 +17,7 @@ def run(cmd, options = {})
     end
 
     def run!(cmd, options = {})
+      AwesomeSpawn.logger = logger
```
Fryguy: This is weird to do it on every call to run.  Not sure where you would do it otherwise though.

brandondunne:
@Fryguy Please review :icecream:

lib/linux_admin/disk.rb
```diff
@@ -5,8 +5,10 @@
 
 require 'linux_admin/partition'
 
-class LinuxAdmin
-  class Disk < LinuxAdmin
+module LinuxAdmin
+  class Disk
+    include Common
```
Fryguy: :tada:

lib/linux_admin/package.rb
```diff
@@ -1,4 +1,7 @@
-class LinuxAdmin
-  class Package < LinuxAdmin
+module LinuxAdmin
+  class Package
+    class << self
+      include Common
+    end
```
Fryguy: This is just `extend Common`

PR #113, opened 2015-02-20
Loosen dependency on inifile

brandondunne:


brandondunne:
@Fryguy Please review

Fryguy:
Looks good to me.   Note that @movitto recently had to use iniparse (as opposed to inifile) for systemd stuff because it supported multiple keys so we may want to switch to that to stop a shared  dependency

brandondunne:
From IRC:
```
<bdunne> Hey, when you get a chance, let me know if you have any concerns with this: https://github.com/ManageIQ/linux_admin/pull/113
<mmorsi> hey
<mmorsi> looks fine to me
<mmorsi> didn't test it tho
<bdunne> all the specs pass
<mmorsi> sure
<mmorsi> we can figure out the inifile / iniparse stuff  in hte future as the systemd progresses
<mmorsi> until then no worries
<bdunne> cool
<bdunne> thanks
<mmorsi> np
```

PR #116, opened 2015-07-14
Use bytes to determine disk size

carbonin:
Changed Disk#size to parse the bytes value directly from the fdisk(8) output.
    
Parsing the converted value was causing an issue as fdisk was using SI units
and the rails methods were expecting binary units.
For example, 5MB was being converted to 5 * 1024 * 1024 bytes by rails
when fdisk intended the value to be 5 * 1000 * 1000 bytes.
This change avoids the confusion by parsing the bytes value directly.
    
https://bugzilla.redhat.com/show_bug.cgi?id=1096331

### DIFF
--- a/lib/linux_admin/disk.rb
+++ b/lib/linux_admin/disk.rb
@@ -70,7 +70,7 @@ module LinuxAdmin
         out = run!(cmd(:fdisk), :params => {"-l" => nil}).output
         out.each_line { |l|
           if l =~ /Disk #{path}: .*B, (\d+) bytes/
-            size = $1.to_f
+            size = $1.to_i
             break
           end
         }
--- a/lib/linux_admin/disk.rb
+++ b/lib/linux_admin/disk.rb
@@ -68,12 +68,12 @@ module LinuxAdmin
       @size ||= begin
         size = nil
         out = run!(cmd(:fdisk), :params => {"-l" => nil}).output
-        out.each_line { |l|
-          if l =~ /Disk #{path}: .*B, (\d+) bytes/
-            size = $1.to_i
+        out.each_line do |l|
+          /Disk #{path}: .*B, (\d+) bytes/.match(l) do |m|
+            size = m[1].to_i
             break
           end
-        }
+        end
         size
       end
     end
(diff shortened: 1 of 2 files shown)


spec/disk_spec.rb
```diff
@@ -37,7 +37,7 @@
 
       disk = LinuxAdmin::Disk.new :path => '/dev/hda'
       allow(disk).to receive(:run!).and_return(double(:output => fdisk))
-      expect(disk.size).to eq(500.1.gigabytes)
+      expect(disk.size).to eq(500107862016)
```
Fryguy: Might be clearer to write out the calculation here to show what that long number represents.  I assume it's `500.1 * 1024 * 1024`

carbonin: No, that's the issue.  The bytes value is the only thing we can rely on from the fdisk output.
This value would technically be exactly either 465.761741638 GiB (500107862016 / 1024^3) or 500.107862016 GB (500107862016 / 1000^3).

The bug was that we were interpreting the GB as GiB

Fryguy: But what's the math to get the 107862016 part?

carbonin: 500107862016 / 1000^3

Fryguy: OHHHHH I totally misunderstood...you changed it to just return bytes, so we just use the value directly.

lib/linux_admin/disk.rb
```diff
@@ -69,8 +69,8 @@ def size
         size = nil
         out = run!(cmd(:fdisk), :params => {"-l" => nil}).output
         out.each_line { |l|
-          if l =~ /Disk #{path}: ([0-9\.]*) ([KMG])B.*/
-            size = str_to_bytes($1, $2)
+          if l =~ /Disk #{path}: .*B, (\d+) bytes/
+            size = $1.to_f
```
Fryguy: Why `.to_f`?  Can you have fractional bytes?  Or is this just to keep the code consistent with returning a float beforehand?

spec/disk_spec.rb
carbonin: Exactly.  We were incorrectly converting to bytes when we were just given the value in bytes to start with so we didn't need to be converting at all.

lib/linux_admin/disk.rb
carbonin: That could be changed.  This method should always return a whole number of bytes.

PR #118, opened 2015-07-24
Normalize service classes

Fryguy:
@carbonin @jrafanie Please review.  Let me know if the name SysvService is good for classic services.  Wasn't sure about that.


jrafanie:
:+1:

PR #120, opened 2015-07-27
[WIP] Add scap lockdown

carbonin:
Exposes access to various SCAP settings through linux admin.  A full list of items can be found on Trello (https://trello.com/c/Zh1TAxfm/50-8-7-1-phase-2-config-scap-stig-security)


Fryguy:
Are these OS specific?  That is, do the file paths only apply to a particular OS?  If so, we might have to put this in a subclass and call it via Distros or something.

carbonin:
Most I would assume are not, I can check a sampling of distros though.  What would be good coverage for that?

lib/linux_admin/security/limits_conf.rb
```diff
@@ -0,0 +1,28 @@
...
+
+      def self.set_value(domain, type, item, value, filename = CONF_FILE)
+        config_text = File.read(filename)
+        new_line = "#{domain} #{type} #{item} #{value}"
+        new_text = config_text.gsub!(/^[^#\n]* #{item} .*/, new_line)
+
+        if new_text
+          File.open(filename, "w") do |file|
+            file.puts(new_text)
+          end
```
Fryguy: This pattern simplifies to just `File.write(filename, new_text)`  If you need the extra new_line from the puts at the end you can append it to new_text manually above

lib/linux_admin/security/limits_conf.rb
```diff
@@ -0,0 +1,28 @@
...
+        else
+          File.open(filename, "a") do |file|
+            file.puts(new_line)
+          end
```
Fryguy: I believe this simplifies to `File.write(filename, new_line, :mode => "a")`

lib/linux_admin/security/modprobe.rb
```diff
@@ -0,0 +1,53 @@
...
+
+        if new_text
+          File.open(filename, "w") do |file|
+            file.puts(new_text)
+          end
+        else
+          File.open(filename, "a") do |file|
+            file.puts(new_line)
+          end
+        end
```
Fryguy: I'm seeing a pattern...this probably should move into a helper method in Common.

lib/linux_admin/security/sysctl_conf.rb
```diff
@@ -0,0 +1,43 @@
...
+
+        if new_text
+          File.open(filename, "w") do |file|
+            file.puts(new_text)
+          end
+        else
+          File.open(filename, "a") do |file|
+            file.puts(new_line)
+          end
+        end
```
Fryguy: Ugh, this file will be written over an over...wonder if we can mitigate that/

lib/linux_admin/security/sysctl_conf.rb
```diff
@@ -0,0 +1,43 @@
...
+        "net.ipv4.conf.all.send_redirects"           => 0
+      }
+
+      def self.apply_scap_settings(filename = CONF_FILE)
+        SCAP_SETTINGS.each do |k, v|
+          set_value(k, v, filename)
+        end
+      end
+
+      def self.set_value(key, val, filename = CONF_FILE)
```
Fryguy: This looks like a private method, so we probably need private here...in fact, we should check ll the new files for public vs private methods.

lib/linux_admin/security/useradd.rb
```diff
@@ -0,0 +1,33 @@
...
+      CONF_FILE = "/etc/default/useradd"
+
+      SCAP_SETTINGS = {
+        "INACTIVE" => 35,
+      }
+
+      def self.apply_scap_settings(filename = CONF_FILE)
+        SCAP_SETTINGS.each do |k, v|
+          set_value(k, v, filename)
+        end
```
Fryguy: Also a common pattern that we can probably extract out.

spec/security/audit_rules_spec.rb
```diff
@@ -0,0 +1,171 @@
...
+  def test_file_contents
+    File.read(test_file_name)
+  end
+
+  around(:each) do |example|
+    text = test_file_contents
+    example.run
+    File.open(test_file_name, "w") do |file|
+      file.puts(text)
+    end
```
Fryguy: Also can be `File.write`
Also, you don't need :each...it's implied as the default.

spec/security/sysctl_conf_spec.rb
```diff
@@ -0,0 +1,78 @@
+
```
Fryguy: Remove leading empty lines (I bet you use VIM ;) )

Fryguy:
> What would be good coverage for that?

`¯\_(ツ)_/¯`

BTW, this is really nice...was this extracted from some other code or a script, or did you have to research this all yourself.

lib/linux_admin/security.rb
```diff
@@ -0,0 +1,18 @@
+module LinuxAdmin
+  class Security
+    def self.scap_lockdown
+      SshdConfig.apply_scap_settings
+      Service.apply_scap_settings
+      SysctlConf.apply_scap_settings
+      LimitsConf.apply_scap_settings
+      Securetty.remove_vcs
```
Fryguy: Should this method be named `apply_scap_settings`?  Same question for AuditRules.

If they are all named the same, this could be written as:

```ruby
[SshdConfig, Service, SysctlConf, LimitsConf, Securetty, LoginDefs, Useradd, AuditRules, AuditRules, Modprobe].each do |klass|
  klass.public_send(:apply_scap_settings)
end
```

or you can multiline the class array if you prefer.

lib/linux_admin/security/modprobe.rb
```diff
@@ -0,0 +1,53 @@
+module LinuxAdmin
+  class Security
+    class Modprobe
+      CONF_FILE_NAME = "/etc/modprobe.d/scap_lockdown"
```
Fryguy: Let's make the constant name consistent across the classes...For example in LoginDefs it's called  CONF_FILE

lib/linux_admin/security/audit_rules.rb
```diff
@@ -0,0 +1,117 @@
...
+          {"arch" => "b64", "exit" => "-EPERM", "auid>" => "500", "auid!" => "4294967295"},
+          "access"
+        ]
+      ]
+
+      def self.apply_scap_settings(filename = CONF_FILE)
+        set_buffer_size(16_384, filename)
+        SCAP_FILESYSTEM_RULES.each do |r|
+          add_filesystem_rule(*r, filename)
+        end
```
Fryguy: Mostly style..but if you make these one-lined it might be cleaner.

lib/linux_admin/security/audit_rules.rb
```diff
@@ -0,0 +1,117 @@
...
+        SCAP_SYSTEM_CALL_RULES.each do |r|
+          add_system_call_rule(*r, filename)
+        end
+      end
+
+      def self.add_filesystem_rule(path, permissions, key_name,
+                                   filename = CONF_FILE)
```
Fryguy: Not a fan of the multi-line params...can they just be one line?

carbonin:
Most of the config file changes were set up in a bash script that I got from when this was brought up the first time a few months ago.  Some stuff was the same, but others had to be changed so it would work with systemd and friends so I had to test/research most of it to be sure it would do what we wanted it to do.  Plus porting the bash stuff to ruby.

lib/linux_admin/security.rb
carbonin: Yeah, I thought about that, but removing access to the vcs didn't seem really SCAP specific.  I could make the method more generic then create an apply_scap_settings method or just have one that calls remove_vcs with the intention of creating other methods to edit this file in the future.

lib/linux_admin/security/common.rb
```diff
@@ -0,0 +1,17 @@
...
+      def set_value(key, val, filename)
+        config_text = File.read(filename)
+        new_line = "#{key} #{val}\n"
+        new_text = config_text.gsub!(/^#*#{key}.*/, new_line)
+
+        if new_text
+          File.write(filename, new_text)
+        else
+          File.write(filename, new_line, :mode => "a")
+        end
```
carbonin: I was thinking of making this whole idea of set_value a bit more generic with something like:

```ruby
def replace_config_line(new_line, rep_regex, file_text)
  new_text = config_text.gsub!(rep_regex, new_line)
  if new_text
    return new_text
  else
    return new_text << new_line
end
```

Then we can also use this multiple times in the calling class and let it handle saving the file using the returned text.  This way we will always overwrite the contents, but only ever do it once.  What do you think @Fryguy ?

carbonin:
@Fryguy What should I do around the gem version for this PR?

Fryguy:
Nothing.  When we release we'll decide on the version number

carbonin:
Looks like everything is where we expect it to be in RHEL 7.1

lib/linux_admin/security/sshd_config.rb
```diff
@@ -0,0 +1,31 @@
...
+        SCAP_SETTINGS.each do |k, v|
+          new_line = "#{k} #{v}\n"
+          config_text = replace_config_line(new_line, /^#*#{k}.*\n/, config_text)
+        end
+        File.write(filename, config_text)
+      end
+
+      def restart_service
+        serv = LinuxAdmin::Service.new("sshd")
+        serv.running? ? serv.restart : serv.start
```
Fryguy: I feel like this logic belongs right inside `service.restart`.  That way the caller just calls `service.restart`, and the Service class takes care of it all.  What do you think?  Doesn't have to be in this PR.

lib/linux_admin/security/audit_rules.rb
```diff
@@ -0,0 +1,112 @@
...
+        ["/etc/issue.net",         "wa", "audit_network_modifications"],
+        ["/etc/hosts",             "wa", "audit_network_modifications"],
+        ["/etc/sysconfig/network", "wa", "audit_network_modifications"]
+      ]
+
+      SCAP_SYSTEM_CALL_RULES = [
+        [
+          "always",
+          "exit",
+          %w(sethostname setdomainname),
```
brandondunne: Sorry, this seems strange to me.  I'd prefer to see the ```[ ]``` syntax instead of the %w here.

lib/linux_admin/security/common.rb
```diff
@@ -0,0 +1,14 @@
+module LinuxAdmin
+  class Security
+    module Common
+      def replace_config_line(new_line, rep_regex, file_text)
+        new_text = file_text.gsub!(rep_regex, new_line)
+        if new_text
```
brandondunne: Prefer:
```ruby
return new_text if new_text
file_text << new_line
```
rather than if/else/end

lib/linux_admin/security/common.rb
```diff
@@ -0,0 +1,14 @@
+module LinuxAdmin
+  class Security
+    module Common
+      def replace_config_line(new_line, rep_regex, file_text)
+        new_text = file_text.gsub!(rep_regex, new_line)
```
brandondunne: I don't think you want the ```!``` after the gsub since it will modify the original "file_text".

lib/linux_admin/security/sshd_config.rb
carbonin: Just did a quick test and it looks like restart will succeed even if the service is stopped.  So I can just take the method out and call to LinuxAdmin::Service directly.

lib/linux_admin/security/audit_rules.rb
carbonin: I have no preference, I did it this was to appease RuboCop.

lib/linux_admin/security/modprobe.rb
```diff
@@ -0,0 +1,37 @@
...
+        new_line = "install #{mod_name} /bin/true\n"
+        new_text = replace_config_line(new_line, /^install #{mod_name}.*\n/, config_text)
+        File.write(filename, new_text)
+      end
+
+      def enable_module(mod_name, filename)
+        begin
+          config_text = File.read(filename)
+        rescue Errno::ENOENT
+          return
```
brandondunne: Don't need the return

spec/security/modprobe_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  describe "#disable_module" do
+    it "adds a module to the file" do
+      described_class.new.disable_module("test_module", test_file_name)
+      expect(test_file_contents).to match(%r{^install test_module /bin/true\n})
+    end
+
+    context "creates a file" do
+      after do
+        File.exist?(new_file_name) && File.delete(new_file_name)
```
brandondunne: You can use [FileUtils.rm or rm_f](http://ruby-doc.org/stdlib-2.2.2/libdoc/fileutils/rdoc/FileUtils.html#method-c-rm) to avoid the existence check

spec/security/modprobe_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+      end
+      it "if the given one doesn't exist" do
+        expect(File.exist?(new_file_name)).to be false
+        described_class.new.disable_module("test_module", new_file_name)
+        expect(File.exist?(new_file_name)).to be true
```
brandondunne: I don't think you need this since the next ```expect``` covers this too, correct?

lib/linux_admin/security/common.rb
carbonin: I do in this case, because the other version will just return a copy of the original string rather than nil if no substitutions were made.

brandondunne: Ok, wasn't sure if you were aware that it would modify the original.

lib/linux_admin/security/modprobe.rb
carbonin: If I take that return out wont we run lines 32 and 33?  If the file doesn't exist, we don't want to do that.

lib/linux_admin/security/common.rb
carbonin: I could do

```ruby
String.new(file_text).gsub!(rep_regex, new_line)
```

because the caller of this method should expect a return, rather than using the passed in object.  I think that will lead to less confusion as I'm only using this version because of the return behavior. What do you think?

carbonin: Although I would still be returning the same string object in the case where there are no substitutions made ... That might be even more confusing.

spec/security/modprobe_spec.rb
carbonin: Technically, yes, but with this line we would get a failed test rather than an exception if the file doesn't get created.

Fryguy:
@carbonin Let's investigate OpenSCAP and see if we should proceed with this PR.

Fryguy:
Closing in favor of #121

PR #121, opened 2015-08-07
Add openscap lockdown

carbonin:
Added the Scap class which, given a list of openscap rule ids and values, will run an evaluation using a temporary profile and fix any failures.

### DIFF
--- a/lib/linux_admin/scap.rb
+++ b/lib/linux_admin/scap.rb
@@ -23,6 +23,8 @@ module LinuxAdmin
       values = args.last.kind_of?(Hash) ? args.pop : {}
       rules = args
 
+      raise "No SCAP rules provided" if rules.empty?
+
       with_xml_files(rules, values) do |xccdf_file_path|
         lockdown_profile(xccdf_file_path, PROFILE_ID)
       end
--- a/lib/linux_admin/scap.rb
+++ b/lib/linux_admin/scap.rb
@@ -1,0 +1,102 @@
+require 'nokogiri'
+
+module LinuxAdmin
+  class Scap
+    PROFILE_ID = "linux-admin-scap"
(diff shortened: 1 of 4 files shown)


lib/linux_admin/scap.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    def run_openscap_fix
+      FileUtils.cp("#{SSG_XML_PATH}#{@oval_file}", '/tmp')
+
+      session = OpenSCAP::Xccdf::Session.new("/tmp/#{@xccdf_file}")
+      session.load
+      session.profile = PROFILE_NAME
+      session.evaluate
+      session.remediate
+      session.destroy
```
carbonin: All of these lines could potentially raise an OpenSCAPError.  Is it better to let the caller of lockdown handle that or should we do something different?

carbonin:
Looks like the tests are failing because openscap isn't installed on the travis machine.  Just requiring the openscap gem needs the package to be installed. @Fryguy How can we fix that?

Both "openscap" and "scap-security-guide" would need to be installed for the actual feature to work.

lib/linux_admin/scap.rb
```diff
@@ -0,0 +1,104 @@
...
+    def with_xml_files(rules, values)
+      FileUtils.cp(oval_file, tempdir)
+
+      Tempfile.create("scap_xccdf") do |f|
+        write_xccdf_xml(f, profile_xml(PROFILE_ID, rules, values))
+        f.close
+        yield f.path
+      end
+    ensure
+      FileUtils.rm_f(tempdir.join(File.basename(oval_file)))
```
kbrock: should this be `FileUtils.rm_f(tempdir.path)`?

PR #122, opened 2015-09-24
Added class to wrap timedatectl command

carbonin:
@Fryguy @bdunne

### DIFF
--- a/lib/linux_admin/time_date.rb
+++ b/lib/linux_admin/time_date.rb
@@ -3,6 +3,8 @@ module LinuxAdmin
     extend Common
     COMMAND = 'timedatectl'
 
+    TimeCommandError = Class.new(StandardError)
+
     def self.system_timezone
       result = run(cmd(COMMAND), :params => ["status"])
       result.output.split("\n").each do |l|
@@ -12,10 +14,14 @@ module LinuxAdmin
 
     def self.system_time=(time)
       run!(cmd(COMMAND), :params => ["set-time", "#{time.strftime("%F %T")}", :adjust_system_clock])
+    rescue AwesomeSpawn::CommandResultError => e
+      raise TimeCommandError, e.message
     end
 
     def self.system_timezone=(zone)
       run!(cmd(COMMAND), :params => ["set-timezone", zone])
+    rescue AwesomeSpawn::CommandResultError => e
+      raise TimeCommandError, e.message
     end
   end
 end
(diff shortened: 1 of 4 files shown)


lib/linux_admin/time_date.rb
```diff
@@ -0,0 +1,21 @@
...
+    COMMAND = 'timedatectl'
+
+    def self.timezone
+      result = run(cmd(COMMAND), :params => ["status"])
+      result.output.split("\n").each do |l|
+        return l.split(':')[1].strip if l =~ /Time.*zone/
+      end
+    end
+
+    def self.set_system_time(time)
```
Fryguy: I think an accessor looking method would be better: `def self.system_time=(time)`

```ruby
LinuxAdmin::TimeDate.system_time = Time.now
```

lib/linux_admin/time_date.rb
```diff
@@ -0,0 +1,21 @@
...
+      run!(cmd(COMMAND), :params => ["set-time", "#{time.strftime("%F %T")}", :adjust_system_clock])
+    end
+
+    def self.set_system_timezone(location, city)
```
Fryguy: If the getter up above is `timezone` should this be `timezone=` (with the combined timezone string passed in?)

spec/time_date_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  describe ".timezone" do
+    it "returns the correct timezone" do
+      awesome_spawn_args = [
+        RUN_COMMAND,
+        :params => ["status"]
+      ]
+      expect(AwesomeSpawn).to receive(:run).with(*awesome_spawn_args).and_return(timedatectl_result)
+      tz = described_class.timezone
+      expect(tz).to eq("America/New_York (EDT, -0400)")
```
Fryguy: I feel like this method should match what you would pass into `timezone=`, though I can understand needing the details.  Perhaps a second method `timezone_detailed` or something?

lib/linux_admin/time_date.rb
carbonin: I kind of like adding system like the `system_time=` method.  I think it communicates that we are altering the underlying system setting rather than just storing data in class variables or something like that.

lib/linux_admin/time_date.rb
```diff
@@ -0,0 +1,21 @@
+module LinuxAdmin
+  class TimeDate
+    extend Common
+    COMMAND = 'timedatectl'
```
carbonin: Would we want to do something like https://github.com/ManageIQ/linux_admin/pull/123#discussion_r40456731 for this too? @Fryguy

kbrock:
:+1: LGTM

for a separate PR, but is there an abstraction for `sysctl` hidden in here somewhere?

PR #123, opened 2015-09-25
Added method to get and set hostname to Hosts class

carbonin:
Added a method for setting the host name via `hostnamectl` to the Hosts class.

### DIFF
--- a/lib/linux_admin/hosts.rb
+++ b/lib/linux_admin/hosts.rb
@@ -39,7 +39,12 @@ module LinuxAdmin
     end
 
     def hostname=(name)
-      run!(cmd('hostnamectl'), :params => ['set-hostname', name])
+      if cmd?("hostnamectl")
+        run!(cmd('hostnamectl'), :params => ['set-hostname', name])
+      else
+        File.write("/etc/hostname", name)
+        run!(cmd('hostname'), :params => {:file => "/etc/hostname"})
+      end
     end
 
     def hostname
--- a/lib/linux_admin/hosts.rb
+++ b/lib/linux_admin/hosts.rb
@@ -42,6 +42,11 @@ module LinuxAdmin
       run!(cmd('hostnamectl'), :params => ['set-hostname', name])
     end
 
+    def hostname
+      result = run(cmd("hostname"))
+      result.success? ? result.output : nil
+    end
+
   private
     def parse_file
       @parsed_file = []
--- a/lib/linux_admin/hosts.rb
+++ b/lib/linux_admin/hosts.rb
@@ -1,5 +1,7 @@
 module LinuxAdmin
   class Hosts
+    include Common
+
     attr_accessor :filename
     attr_accessor :raw_lines
     attr_accessor :parsed_file
@@ -34,6 +36,10 @@ module LinuxAdmin
         @parsed_file.store_path(line_number, :hosts, new_hosts)
         @parsed_file.store_path(line_number, :comment, comment) if comment
       end
+    end
+
+    def hostname=(name)
+      run!(cmd('hostnamectl'), :params => ['set-hostname', name])
     end
 
   private
--- a/spec/hosts_spec.rb
+++ b/spec/hosts_spec.rb
@@ -1,4 +1,5 @@
 describe LinuxAdmin::Hosts do
+  TEST_HOSTNAME = "test-hostname"
   etc_hosts = "\n #Some Comment\n127.0.0.1\tlocalhost localhost.localdomain # with a comment\n127.0.1.1  my.domain.local"
   before do
     allow(File).to receive(:read).and_return(etc_hosts)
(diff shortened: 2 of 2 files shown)


lib/linux_admin/hosts.rb
```diff
@@ -36,6 +38,10 @@ def update_entry(address, hostname, comment = nil)
       end
     end
 
+    def hostname=(name)
+      run!(cmd('hostnamectl'), :params => ['set-hostname', name])
```
Fryguy: Is all of LinuxAdmin systemctl based?  If not, can we make this so that it does it properly depending on the system?  @bdunne Thoughts on organizing this?

carbonin: So traditionally `/etc/hostname` and the `hostname` command were used to deal with this.

With systemd systems we got the new `systemd-hostnamed` service which seems to be adding a layer of complexity to this.

For example:
I would have expected running `hostnamectl set-hostname new-host` to have the same effect as `echo "new-host" > /etc/hostname && hostname -F /etc/hostname`, but after running the latter commands `hostnamectl status` still shows the old name and doesn't change until I run `systemctl restart systemd-hostnamed`

So I'm thinking like this:

```ruby
if cmd?("hostnamectl")
  run!(cmd('hostnamectl'), :params => ['set-hostname', name])
else
  File.write("/etc/hostname", name)
  run!(cmd('hostname'), :params => ["-F", "/etc/hostname"])
end
```
What do you think @Fryguy ?

Fryguy: Yeah, that sounds good to me...We might want a higher level check for systemd instead of the individual commands, but this should work.

Also the params on the second one can be simplified to `:params => {:F => "/etc/hostname}`

carbonin: Yeah, it's `--file` so I'll use `:params => {:file => "/etc/hostname"}`

But on a side note:
```ruby
irb(main):004:0> AwesomeSpawn::CommandLineBuilder.new.build("hostname", {:F => "/etc/hostname"})
=> "hostname --F /etc/hostname"
```
Am I missing something?

lib/linux_admin/hosts.rb
```diff
@@ -36,6 +38,20 @@ def update_entry(address, hostname, comment = nil)
...
+      if cmd?("hostnamectl")
+        run!(cmd('hostnamectl'), :params => ['set-hostname', name])
+      else
+        File.write("/etc/hostname", name)
+        run!(cmd('hostname'), :params => {:file => "/etc/hostname"})
+      end
+    end
+
+    def hostname
+      result = run(cmd("hostname"))
```
bdunne: We already require `activesupport/core_ext` so you could simplify this if there is nothing on output when unsuccessful.
```ruby
run(cmd("hostname")).output.presence
```

bdunne:
:+1: looks good, just the one comment.

lib/linux_admin/hosts.rb
carbonin: I'm not sure about what will be on output.  It depends really on what `Open3.capture3` returns here (https://github.com/ManageIQ/awesome_spawn/blob/master/lib/awesome_spawn.rb#L119) and what `hostname` returns in whatever failure case happened.

kbrock:
This is nice.

I think being explicit around `success?` is nice. probably overkill but... kinda like it.

:+1: LGTM

lib/linux_admin/hosts.rb
kbrock: oh noes. That is a bug. [single letter](https://github.com/ManageIQ/awesome_spawn/commit/2a03df8aa0567a63332512bfb7fa427ddc003d0e)

lib/linux_admin/hosts.rb
```diff
@@ -36,6 +38,20 @@ def update_entry(address, hostname, comment = nil)
       end
     end
 
+    def hostname=(name)
```
carbonin: Do we want to add the logic to add the machine's IP address to the `/etc/hosts` file (https://github.com/ManageIQ/manageiq-appliance/blob/master/LINK/usr/bin/miqnet.sh#L167) here or have the caller do it in a separate call?
@Fryguy @bdunne

spec/hosts_spec.rb
```diff
@@ -50,4 +50,52 @@
...
+        @instance.cmd('hostnamectl'),
+        :params => ['set-hostname', new_hostname]
+      ]
+      expect(@instance).to receive(:cmd?).with("hostnamectl").and_return(true)
+      expect(AwesomeSpawn).to receive(:run!).with(*spawn_args)
+      @instance.hostname = new_hostname
+    end
+
+    it "sets the hostname with hostname when hostnamectl does not exist" do
+      new_hostname = "test-hostname"
```
gtanzillo: Put new_hostname and hostname below in a common place.

spec/hosts_spec.rb
```diff
@@ -50,4 +50,52 @@
...
+      expect(File).to receive(:write).with("/etc/hostname", new_hostname)
+      expect(AwesomeSpawn).to receive(:run!).with(*spawn_args)
+      @instance.hostname = new_hostname
+    end
+  end
+
+  describe "#hostname" do
+    it "returns the hostname" do
+      hostname = "test-host"
+      spawn_args = [
```
gtanzillo: DRY up spawn_args in a let or before

lib/linux_admin/hosts.rb
kbrock: If you are setting ip address, would you ensure that if you set a different ip address, it clears out the old ip address from `/etc/hosts`
If you are changing the hostname, do we also clear that out.

man, the use cases around this are complicated.

carbonin: I would say that the safest way to go would be to do as little as possible with each call.  I don't really want to assume what people will want to do with these tools.  I think it will be more clear from the calling code to do something like this https://github.com/ManageIQ/manageiq/pull/4558/files#diff-16bd2bb3c0abfe4c0381896614e7b987R238

PR #124, opened 2015-09-25
Added class for managing IP addresses

carbonin:
Added a class that currently allows retrieving IP addresses via `hostname -I` using a retry mechanism.

### DIFF
--- a/lib/linux_admin/ip_address.rb
+++ b/lib/linux_admin/ip_address.rb
@@ -1,0 +1,27 @@
+module LinuxAdmin
+  class IpAddress
+    include Common
+
+    def address
+      address_list.detect { |ip| IPAddr.new(ip).ipv4? }
+    end
+
+    def address6
+      address_list.detect { |ip| IPAddr.new(ip).ipv6? }
+    end
+
+    private
+
+    def address_list
+      result = nil
+      # Added retry to account for slow DHCP not assigning an IP quickly at boot; specifically:
+      # https://github.com/ManageIQ/manageiq-appliance/commit/160d8ccbfbfd617bdb5445e56cdab66b9323b15b
+      5.times do
+        result = run(cmd("hostname"), :params => ["-I"])
+        break if result.success?
+      end
+
+      result.success? ? result.output.split(' ') : []
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


lib/linux_admin/ip_addr.rb
```diff
@@ -0,0 +1,31 @@
+module LinuxAdmin
+  class IpAddr
```
Fryguy: I would not name this so similarly to the Ruby class.  Maybe just spell it out (`IpAddress`)?  Should it be part of a larger class like `NetworkConfig` or something?

lib/linux_admin/ip_addr.rb
```diff
@@ -0,0 +1,31 @@
...
+    include Common
+
+    def ip4
+      address_list.each do |ip|
+        return ip if IPAddr.new(ip).ipv4?
+      end
+      nil
```
Fryguy: You can simplify this with detect

```ruby
def ip4
  address_list.detect { |ip| IPAddr.new(ip).ipv4? }
end
```

lib/linux_admin/ip_addr.rb
```diff
@@ -0,0 +1,31 @@
+module LinuxAdmin
+  class IpAddr
+    include Common
+
+    def ip4
```
Fryguy: Should the method be named `ipv4` since that's how people would expect it to be named?

lib/linux_admin/ip_addr.rb
carbonin: @Fryguy If we go the route of creating a larger class we would be able to add stuff like DNS servers to it rather than having a separate class for that, but in that case I would also want to move the functionality from https://github.com/ManageIQ/linux_admin/blob/master/lib/linux_admin/hosts.rb to the larger class.  Is that a change we want to make to the API?

lib/linux_admin/ip_address.rb
```diff
@@ -0,0 +1,25 @@
+module LinuxAdmin
+  class IpAddress
+    include Common
+
+    def ipv4
```
gtanzillo: Would be good to rename to address and address6 to be consistent with system utilities.

lib/linux_admin/ip_address.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    def ipv6
+      address_list.detect { |ip| IPAddr.new(ip).ipv6? }
+    end
+
+    private
+
+    def address_list
+      result = nil
+      5.times do
```
gtanzillo: Add a comment or link to PR that explains why 5.times is necessary.

PR #125, opened 2015-10-02
Fix NameError on IPAddr when running tests locally

jrafanie:
Introduced in commit: 17c5d4434971e, PR #124

Not quite sure why the test suite didn't fail on travis but it was failing for me locally.


jrafanie:
cc @carbonin

carbonin:
:+1:

PR #127, opened 2015-10-05
Added Dns class for managing DNS search and nameservers

carbonin:
@bdunne @Fryguy Please review

https://trello.com/c/rWKh4KQs


lib/linux_admin/dns.rb
```diff
@@ -0,0 +1,41 @@
...
+    attr_accessor :nameservers
+    attr_accessor :search_order
+
+    def initialize(filename = "/etc/resolv.conf")
+      @filename = filename
+      reload
+    end
+
+    def reload
+      parse_file(File.read(@filename).split("\n"))
```
bdunne: The split feels like it belongs in `#parse_file`

bdunne: I would expect `#reload` to clear the `attr_accessors`

lib/linux_admin/dns.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    def save
+      contents = "search #{@search_order.join(" ")}\n"
+      @nameservers.each { |ns| contents += "nameserver #{ns}\n" }
+      File.write(@filename, contents)
+    end
+
+    private
+
+    def parse_file(lines)
```
bdunne: I think the `File.read` belongs in this method and it shouldn't accept any params.

lib/linux_admin/dns.rb
carbonin: `parse_file` clears them to start.  Would it look better to clear them first here?

spec/dns_spec.rb
```diff
@@ -0,0 +1,111 @@
+describe LinuxAdmin::Dns do
+  SEARCH_ORDER = %w(ewr.redhat.com ewr.redhat.com. redhat.com. manageiq.com)
+  NAMESERVERS  = %w(10.11.5.19 10.5.30.160 192.168.252.3)
+
+  NEW_ORDER = %w(thing.com other.com)
+  NEW_NS    = %w(1.2.3.4 5.6.7.8)
+
+  def dns_file_path
```
bdunne: You can use `let` here

spec/dns_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+
+  describe "#reload" do
+    NEW_CONTENTS = <<DNS_END
+search thing.com other.com
+nameserver 1.2.3.4
+nameserver 5.6.7.8
+DNS_END
+
+    def change_file
+      File.write(dns_file_path, NEW_CONTENTS)
```
bdunne: We should try to avoid writing files in a spec, especially files that hold test data.  In the spec, you should be able to:
```ruby
allow(File).to receive(:read).and_return(original_content, new_content)
```

spec/dns_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+      expect(dns.search_order).to eq(SEARCH_ORDER)
+
+      change_file
+      dns.reload
+
+      expect(dns.search_order).to eq(NEW_ORDER)
+    end
+  end
+
+  describe "#save" do
```
bdunne: Don't let these tests overwrite the test data, instead:
```ruby
expect(File).to receive(:write).with(expected_string)
```

spec/data/dns/resolv.conf
```diff
@@ -0,0 +1,5 @@
+# Generated by NetworkManager
```
bdunne: You can probably put all of this content into a heredoc in the spec file since it is so small.

spec/dns_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+    it "adds a search host" do
+      new_host = "new.server.com"
+      new_search_order = "search #{SEARCH_ORDER.join(' ')} #{new_host}"
+
+      expect(File).to receive(:write) do |file, contents|
+        expect(file).to eq(dns.filename)
+        expect(contents.include?(new_search_order)).to be true
```
bdunne: You can `expect(contents).to include(new_search_order)`

spec/dns_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+    end
+
+    it "replaces the search order" do
+      old_search_order = "search #{SEARCH_ORDER.join(' ')}"
+      new_search_order = "search #{NEW_ORDER.join(' ')}"
+
+      expect(File).to receive(:write) do |file, contents|
+        expect(file).to eq(dns.filename)
+        expect(contents.include?(old_search_order)).to be false
+        expect(contents.include?(new_search_order)).to be true
```
bdunne: You don't need to do both the false and true here if you can compare them using eq()

spec/dns_spec.rb
bdunne: Since the save method writes the entire file, you don't really need multiple tests for search lines and nameserver lines.  You can just test that given these things in the these things in the attrs, here is the exact content that should be written to this file

spec/dns_spec.rb
```diff
@@ -0,0 +1,118 @@
+describe LinuxAdmin::Dns do
+  RESOLV_CONF = <<DNS_END
+# Generated by NetworkManager
+search test.one.com test.two.com. test.com.
```
bdunne: Please don't use public domains and addresses.  Prefer those reserved for testing and documentation.
Domains: `example.com` and `example.org` and sub domains.
IP addresses: `10.*.*.*`, `172.16.*.* - 172.18.*.*` and `192.168.*.*`

lib/linux_admin/dns.rb
```diff
@@ -0,0 +1,38 @@
...
+    def initialize(filename = "/etc/resolv.conf")
+      @filename = filename
+      reload
+    end
+
+    def reload
+      @search_order = []
+      @nameservers  = []
+      lines = File.read(@filename).split("\n")
+      lines.each do |line|
```
bdunne: lines isn't really needed, the `.each` can go right after the `split`

lib/linux_admin/dns.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    def save
+      contents = "search #{@search_order.join(" ")}\n"
+      @nameservers.each { |ns| contents += "nameserver #{ns}\n" }
+      File.write(@filename, contents)
+    end
+
+    private
+
+    def add_search(line)
```
bdunne: Do you think this would be cleaner?
```ruby
@search_order += line.partition(/^search\s+/)[2].split
```

If that's the case, it could live on line 18.

spec/dns_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+
+  SEARCH_ORDER = %w(test.one.com test.two.com. test.com.)
+  NAMESERVERS  = %w(192.168.1.2 10.10.1.2 192.168.252.3)
+
+  NEW_ORDER = %w(thing.com other.com)
+  NEW_NS    = %w(1.2.3.4 5.6.7.8)
+
+  FILE = "/etc/dns_file"
+
+  let(:dns) do
```
bdunne: I feel like this should this be `subject do`

bdunne:
@carbonin Sorry for all the comments, but I really like where this is going. :+1:

spec/dns_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+      expect(subject.nameservers).to eq(NAMESERVERS)
+    end
+
+    it "parses the search order" do
+      expect(subject.search_order).to eq(SEARCH_ORDER)
+    end
+  end
+
+  describe "#reload" do
+    it "reloads the nameservers" do
```
bdunne: These two tests could also be merged in a followup PR

spec/dns_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  FILE = "/etc/dns_file"
+
+  subject do
+    allow(File).to receive(:read).and_return(RESOLV_CONF)
+    described_class.new(FILE)
+  end
+
+  describe ".new" do
+    it "sets the filename" do
```
bdunne: These two tests could also be merged in a followup PR

bdunne:
:+1: LGTM, Thanks!

PR #128, opened 2015-10-07
Fix warnings around Singleton instance method redefinition

carbonin:
Calling .__init__ on Singleton directly was redefining the .instance method
for the singleton class.
This change calls the private method #refresh on the FSTab instance rather
than attempting to reinitialize the instance.

### DIFF
--- a/spec/fstab_spec.rb
+++ b/spec/fstab_spec.rb
@@ -1,10 +1,7 @@
 require 'stringio'
 
 describe LinuxAdmin::FSTab do
-  before do
-    # Reset the singleton so subsequent tests get a new instance
-    Singleton.send :__init__, LinuxAdmin::FSTab
-  end
+  subject { described_class.dup }
 
   it "has newline, single spaces, tab" do
     fstab = <<eos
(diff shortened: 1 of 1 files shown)


Fryguy:
Should the clearing of the singleton cache be done higher as it was previously but just in a different way?

carbonin:
The way it was done before didn't actually call the `#refresh` method so it didn't need to be stubbed out.  If I added it as a `before` I would need to stub there also and because I might be using `instance` for the first time I would need to stub multiple calls which could eat up the calls to `File.read` we are `expect`ing in the tests.

spec/fstab_spec.rb
```diff
@@ -1,18 +1,14 @@
...
-  end
-
   it "has newline, single spaces, tab" do
     fstab = <<eos
 
   
 	
 eos
-    expect(File).to receive(:read).with('/etc/fstab').and_return(fstab)
+    expect(File).to receive(:read).at_least(:once).with('/etc/fstab').and_return(fstab)
```
bdunne: This should only be read once, right?

spec/fstab_spec.rb
```diff
@@ -1,18 +1,14 @@
 require 'stringio'
 
 describe LinuxAdmin::FSTab do
-  before do
```
bdunne: @carbonin I think this before should be changed to `subject { described_class.dup }`.  Than you can just change `LinuxAdmin::FSTab` to `subject` and the warnings should be gone.

spec/fstab_spec.rb
carbonin: That was to account for if it was the first time we referenced instance and also when we call refresh

bdunne:
:+1:

PR #132, opened 2015-10-12
Add distro specific network interface class

carbonin:
Implements the current functions of the LinuxAdmin::IpAddress class as well as distro specific functions for setting IP addresses and DNS servers.

Uses the distro detection in LinuxAdmin::Distros to determine which sub-class to instantiate at run-time.

Currently supports Red Hat based distros (CentOS, Fedora, and RHEL).

@bdunne @Fryguy please review.


lib/linux_admin/network_interface.rb
```diff
@@ -0,0 +1,138 @@
...
+    # Detects the local linux distro and returns the associated NetworkInterface subclass
+    #
+    # @return [Class] The NetworkInterface subclass associated with the local distro
+    def self.dist_class_uncached
+      if [Distros.rhel, Distros.fedora].include?(Distros.local)
+        return NetworkInterfaceRH
+      else
+        return NetworkInterfaceGeneric
+      end
+      raise "No matching class for distro #{Distros.local}"
```
bdunne: You'll never hit this raise, right?

lib/linux_admin/network_interface.rb
```diff
@@ -0,0 +1,138 @@
...
+    include Common
+
+    # Cached class instance variable for what distro we are running on
+    @dist_class = nil
+
+    # Gets the subclass specific to the local Linux distro
+    #
+    # @param reload [Boolean] Determines if the cached value will be reloaded
+    # @return [Class] The proper class to be used
+    def self.dist_class(reload = false)
```
bdunne: Typically we do something like:
```ruby
@dist_class = nil if reload
@dist_class ||= begin
do_some_stuff
end
```

lib/linux_admin/network_interface.rb
```diff
@@ -0,0 +1,138 @@
...
+    # @param interface [String] Name of the network interface to manage
+    def initialize(interface)
+      @interface = interface
+    end
+
+    # Retrieve the IPv4 address assigned to the interface
+    #
+    # @return [String] IPv4 address for the managed interface
+    def address
+      return nil unless (ip_output = ip_show)
```
bdunne: You shouldn't need the `nil` there.

lib/linux_admin/network_interface.rb
```diff
@@ -0,0 +1,138 @@
...
+
+      parse_ip_output(result.output, /^default/, 2)
+    end
+
+    private
+
+    # Detects the local linux distro and returns the associated NetworkInterface subclass
+    #
+    # @return [Class] The NetworkInterface subclass associated with the local distro
+    def self.dist_class_uncached
```
bdunne: I think this code should live inside `self.dist_class`.

lib/linux_admin/network_interface.rb
```diff
@@ -0,0 +1,138 @@
...
+    end
+    private_class_method :dist_class_uncached
+
+    # Parses the output of `ip addr show`
+    #
+    # @param output [String] The command output
+    # @param regex  [Regexp] Regular expression to match the desired output line
+    # @param col    [FixNum] The whitespace delimited column to be returned
+    # @return [String] The parsed data
+    def parse_ip_output(output, regex, col)
```
bdunne: Should we document private methods?

lib/linux_admin/network_interface.rb
```diff
@@ -0,0 +1,138 @@
...
+      the_line = output.split("\n").detect { |l| l =~ regex }
+      the_line.nil? ? nil : the_line.strip.split(' ')[col]
+    end
+
+    # Runs the command `ip addr show <interface>`
+    #
+    # @return [String] The command output, nil on failure
+    def ip_show
+      result = run(cmd("ip"), :params => ["addr", "show", @interface])
+      result.success? ? result.output : nil
```
bdunne: What is in `result.output` on failure?

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    # Parses the interface configuration file into the @interface_conf hash
+    def reload
+      @interface_conf = {}
+      contents = File.read(@iface_file)
+
+      contents.split("\n").each do |line|
+        next if line =~ /^\s*#/
+
+        pair = line.split('=')
```
bdunne: The ifcfg-* files sometimes contain spaces around the =

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,106 @@
...
+    # @param [String] Name of the network interface to manage
+    def initialize(interface)
+      super
+      @iface_file = Pathname.new(IFACE_DIR).join("ifcfg-#{@interface}")
+      reload
+    end
+
+    # Parses the interface configuration file into the @interface_conf hash
+    def reload
+      @interface_conf = {}
```
bdunne: Should this have an `attr_reader`?

lib/linux_admin/network_interface.rb
carbonin: Right!  That was from before I put the generic class in.

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,106 @@
...
+      end
+      save_file
+    end
+
+    # Set up the interface to use DHCP
+    # Removes any previously set static networking information
+    def set_dhcp
+      @interface_conf["BOOTPROTO"] = "dhcp"
+      @interface_conf.delete("IPADDR")
+      @interface_conf.delete("NETMASK")
```
bdunne: An alternative to "NETMASK" is "PREFIX" which contains the CIDR version of the subnet mask.  You'll want to ensure that is also gone.

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,106 @@
...
+      @interface_conf.delete("GATEWAY")
+      save_file
+    end
+
+    # Sets one or both DNS servers for this network interface
+    #
+    # @param dns1 [String] The first DNS server
+    # @param dns2 [String] The second DNS server
+    def set_dns(dns1, dns2 = nil)
```
bdunne: You might be better with `*dns` and allowing as many as the user wants to supply.

lib/linux_admin/network_interface.rb
carbonin: It doesn't show up in the generated html docs, but I added it for the developers.

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,106 @@
...
+      @interface_conf["DOMAIN"] = "\"#{domains.join(' ')}\""
+      save_file
+    end
+
+    private
+
+    # Writes the contents of @interface_conf to @iface_file as `key`=`value` pairs
+    def save_file
+      contents = ""
+      @interface_conf.each { |k, v| contents << "#{k}=#{v}\n" }
```
bdunne: Any concern of keys with blank values?  You may want to:
```ruby
@interface_conf.delete_blanks.each { |k, v| contents << "#{k}=#{v}\n" }
```

lib/linux_admin/network_interface/network_interface_rh.rb
carbonin: It seems like only two are supported.

https://www.centos.org/docs/5/html/Deployment_Guide-en-US/s1-networkscripts-interfaces.html
https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Networking_Guide/sec-Using_the_Command_Line_Interface.html

lib/linux_admin/network_interface.rb
carbonin: I'm fine with that.  I borrowed most of this mechanism from https://github.com/ManageIQ/linux_admin/blob/master/lib/linux_admin/service.rb

lib/linux_admin/network_interface.rb
carbonin: Whatever was printed to stdout.

lib/linux_admin/network_interface/network_interface_rh.rb
carbonin: Probably not considering we want to abstract as much of the config file itself as possible.

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,107 @@
...
+    # @param domains [Array<String>] the list of search domains
+    def search_order=(domains)
+      @interface_conf["DOMAIN"] = "\"#{domains.join(' ')}\""
+      save_file
+    end
+
+    private
+
+    # Writes the contents of @interface_conf to @iface_file as `key`=`value` pairs
+    def save_file
```
bdunne: This could be save and be public.

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,107 @@
...
+      contents = ""
+      @interface_conf.delete_blanks.each { |k, v| contents << "#{k}=#{v}\n" }
```
bdunne: You can collect here if you want.

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,107 @@
...
+      @interface_conf["DNS2"] = dns2 unless dns2.nil?
+      save_file
+    end
+
+    # Sets the search domain list for this network interface
+    #
+    # @param domains [Array<String>] the list of search domains
+    def search_order=(domains)
+      @interface_conf["DOMAIN"] = "\"#{domains.join(' ')}\""
+      save_file
```
bdunne: Maybe we should save once instead of each time we set something

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,107 @@
...
+    #
+    # @param address [IPAddr]
+    # @param netmask [IPAddr]
+    # @param gateway [IPAddr]
+    def set_v4_address(address, netmask, gateway)
+      @interface_conf["BOOTPROTO"]     = "static"
+      @interface_conf["IPADDR"]        = address.to_s
+      @interface_conf["NETMASK"]       = netmask.to_s
+      @interface_conf["GATEWAY"]       = gateway.to_s
+      @interface_conf["NM_CONTROLLED"] = "no"
```
bdunne: Should we always disable "NM_CONTROLLED"? maybe it should be moved to save.

lib/linux_admin/network_interface/network_interface_rh.rb
bdunne: Maybe `dns=` for the method name?

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,107 @@
...
+        set_v4_address(address_ip, netmask_ip, gateway_ip)
+      else
+        raise ArgumentError, "Incompatible address formats provided for #{__method__}"
+      end
+      save_file
+    end
+
+    # Set up the interface to use DHCP
+    # Removes any previously set static networking information
+    def set_dhcp
```
bdunne: Method name 'enable_dhcp' would be better

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,107 @@
...
+      @iface_file = Pathname.new(IFACE_DIR).join("ifcfg-#{@interface}")
+      reload
+    end
+
+    # Parses the interface configuration file into the @interface_conf hash
+    def reload
+      @interface_conf = {}
+      contents = File.read(@iface_file)
+
+      contents.split("\n").each do |line|
```
bdunne: You could use `each_line`

spec/network_interface/network_interface_generic_spec.rb
```diff
@@ -0,0 +1,114 @@
+describe LinuxAdmin::NetworkInterfaceGeneric do
```
bdunne: This should really be testing the base class, right?

You could instantiate one of the child classes to test the instance methods.

lib/linux_admin/network_interface/network_interface_rh.rb
carbonin: Just tried this out and we only ever get `DNS1` and `DNS2` written to resolv.conf.  I'll still probably use an array  and `dns=` and just take the first two.

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,101 @@
...
+    # Sets the search domain list for this network interface
+    #
+    # @param domains [Array<String>] the list of search domains
+    def search_order=(domains)
+      @interface_conf["DOMAIN"] = "\"#{domains.join(' ')}\""
+    end
+
+    # Set up the interface to use DHCP
+    # Removes any previously set static networking information
+    def enable_dhcp
```
carbonin: @bdunne Should we also clear the DNS server and search order here and assume that will be set by DHCP?

lib/linux_admin/network_interface.rb
```diff
@@ -0,0 +1,169 @@
...
+    include Common
+
+    # Cached class instance variable for what distro we are running on
+    @dist_class = nil
+
+    # Gets the subclass specific to the local Linux distro
+    #
+    # @param test_dist [Boolean] Determines if the cached value will be reevaluated
+    # @return [Class] The proper class to be used
+    def self.dist_class(test_dist = false)
```
bdunne: `clear_cache` would be better

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,104 @@
...
+    # @param interface [String] Name of the network interface to manage
+    def initialize(interface)
+      super
+      @iface_file = Pathname.new(IFACE_DIR).join("ifcfg-#{@interface}")
+      parse_conf
+    end
+
+    # Parses the interface configuration file into the @interface_conf hash
+    def parse_conf
+      @interface_conf = {"NM_CONTROLLED" => "no"}
```
bdunne: The "NM_CONTROLLED" should probably go at the end in case it is present in the file.

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,104 @@
...
+      contents = File.read(@iface_file)
+
+      contents.each_line do |line|
+        next if line =~ /^\s*#/
+
+        pair = line.split('=').collect(&:strip)
```
bdunne: pair could be "key, value"

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,104 @@
...
+    # @raise ArgumentError if the mask is not formatted properly
+    def netmask=(mask)
+      validate_ip(mask)
+      @interface_conf["NETMASK"] = mask
+    end
+
+    # Sets one or both DNS servers for this network interface
+    #
+    # @param servers Array<String> The DNS servers
+    def dns=(servers)
```
bdunne: the argument can be `*servers` to allow for any number of arguments rather than requiring you to pass an array

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,104 @@
...
+      @interface_conf["DNS1"] = servers[0]
+      @interface_conf["DNS2"] = servers[1] if servers[1]
+    end
+
+    # Sets the search domain list for this network interface
+    #
+    # @param domains [Array<String>] the list of search domains
+    def search_order=(domains)
```
bdunne: same `*domains`

lib/linux_admin/network_interface/network_interface_rh.rb
bdunne: yes

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,104 @@
+require 'ipaddr'
+require 'pathname'
+
+module LinuxAdmin
+  class NetworkInterfaceRH < NetworkInterface
```
chessbyte: What is the `RH` in `NetworkInterfaceRH`?  Does that stand for Red Hat (the company) or RHEL (the OS)?  If the latter, then I would rename to `NetworkInterfaceRHEL`.

lib/linux_admin/network_interface.rb
```diff
@@ -0,0 +1,169 @@
...
+    @dist_class = nil
+
+    # Gets the subclass specific to the local Linux distro
+    #
+    # @param clear_cache [Boolean] Determines if the cached value will be reevaluated
+    # @return [Class] The proper class to be used
+    def self.dist_class(clear_cache = false)
+      @dist_class = nil if clear_cache
+      @dist_class ||= begin
+        if [Distros.rhel, Distros.fedora].include?(Distros.local)
```
bdunne: Does Distros.rhel include centos?

carbonin: Yes, https://github.com/carbonin/linux_admin/blob/master/lib/linux_admin/distro.rb#L15

lib/linux_admin/network_interface/network_interface_rh.rb
carbonin: My intention was "Red Hat" as the class is able to provide network configuration for CentOS, Fedora, and RHEL.

lib/linux_admin/network_interface.rb
```diff
@@ -0,0 +1,183 @@
+require 'ipaddr'
+
+module LinuxAdmin
+  class NetworkInterface
+    include Common
+
+    # Cached class instance variable for what distro we are running on
+    @dist_class = nil
```
Fryguy: This initialization is not necessary.

lib/linux_admin/network_interface.rb
```diff
@@ -0,0 +1,183 @@
...
+
+      alias_method :orig_new, :new
+    end
+
+    # Creates an instance of the correct NetworkInterface subclass for the local distro
+    def self.new(*args)
+      if self == LinuxAdmin::NetworkInterface
+        dist_class.new(*args)
+      else
+        orig_new(*args)
```
Fryguy: No need for the alias_method, just call `super`.

lib/linux_admin/network_interface.rb
```diff
@@ -0,0 +1,183 @@
...
+      end
+    end
+
+    # @return [String] the interface for networking operations
+    attr_reader :interface
```
Fryguy: For an attr, you don't need the `@return [String]` syntax for YARD (I don't think)

lib/linux_admin/network_interface.rb
```diff
@@ -0,0 +1,183 @@
...
+    def address
+      @network_conf[:address]
+    end
+
+    # Retrieve the IPv6 address assigned to the interface
+    #
+    # @return [String] IPv6 address for the managed interface
+    # @raise [ArgumentError] if the given scope is not `:global` or `:link`
+    def address6(scope = :global)
+      if scope == :global
```
Fryguy: Prefer `case scope` over the `if/elsif/else`

lib/linux_admin/network_interface.rb
```diff
@@ -0,0 +1,183 @@
...
+    #
+    # @return [String] IPv4 gateway address
+    def gateway
+      @network_conf[:gateway]
+    end
+
+    # Brings up the network interface
+    #
+    # @return [Boolean] whether the command succeeded or not
+    def ifup
```
Fryguy: Let's get a nicer, human, method name, maybe `start` / `stop`?

lib/linux_admin/network_interface.rb
```diff
@@ -0,0 +1,183 @@
...
+    #
+    # @return [String] the MAC address
+    def mac_address
+      @network_conf[:mac]
+    end
+
+    # Retrieve the IPv4 sub-net mask assigned to the interface
+    #
+    # @return [String] IPv4 netmask
+    def netmask
```
Fryguy: Should `netmask` be `net_mask` or perhaps just `mask`?

lib/linux_admin/network_interface.rb
```diff
@@ -0,0 +1,183 @@
...
+      run(cmd("ifdown"), :params => [@interface]).success?
+    end
+
+    private
+
+    # Parses the output of `ip addr show`
+    #
+    # @param output [String] The command output
+    # @param regex  [Regexp] Regular expression to match the desired output line
+    # @param col    [FixNum] The whitespace delimited column to be returned
```
Fryguy: Fixnum should have a lowercase `n`

lib/linux_admin/network_interface.rb
Fryguy: Are you expecting this command to fail?  Can you just use `run!(stuff).output`?

lib/linux_admin/network_interface.rb
```diff
@@ -0,0 +1,183 @@
...
+      return unless cidr_ip
+
+      parts = cidr_ip.split('/')
+      @network_conf["address6_#{scope}".to_sym] = parts[0]
+      @network_conf["mask6_#{scope}".to_sym] = mask_addr.mask(parts[1]).to_s
+    end
+  end
+end
+
+Dir.glob(File.join(File.dirname(__FILE__), "network_interface", "*.rb")).each { |f| require f }
```
Fryguy: `File.dirname(__FILE__)` can be replaced with `__dir__` (or do we still support old rubies?)

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,138 @@
...
+      super
+      @iface_file = Pathname.new(IFACE_DIR).join("ifcfg-#{@interface}")
+      parse_conf
+    end
+
+    # Parses the interface configuration file into the @interface_conf hash
+    def parse_conf
+      @interface_conf = {}
+
+      File.read(@iface_file).each_line do |line|
```
Fryguy: You can do `File.foreach(@iface_file) do |line|`  Then you aren't reading the whole file into memory at once.

lib/linux_admin/network_interface/network_interface_rh.rb
Fryguy: @bdunne However, this is a setter, so unless it's called with .send, it's not likely they *won't* be passing an Array...even so, I agree, and the code could look like:

```ruby
def dns=(*servers)
  servers = servers.flatten
  # stuff
end
```

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,138 @@
...
+      @interface_conf.delete("DOMAIN")
+    end
+
+    # Applies the given static network configuration to the interface
+    #
+    # @param ip [String] IPv4 address
+    # @param mask [String] subnet mask
+    # @param gw [String] gateway address
+    # @param dns1 [Array<String>] list of dns servers
+    # @param search [Array<String>] list of search domains
```
Fryguy: Declare in the documentation that this is an "Optional list of search domains"

lib/linux_admin/network_interface/network_interface_rh.rb
Fryguy: Incidentally, that also allows one to call `network_interface.dns = "dns_server1"` if they only have 1

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,138 @@
...
+
+    private
+
+    # Validate that the given IPv4 address is formatted correctly
+    #
+    # @param ip [String]
+    # @raise ArgumentError if the address is not correctly formatted
+    def validate_ip(ip)
+      IPAddr.new(ip)
+    rescue
```
Fryguy: Does IPAddr raise a RuntimeError or something specific?  Can you catch the specific error only?

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,138 @@
...
+      raise ArgumentError, "#{ip} is not a valid IPv4 address"
```
Fryguy: What if it's IPv6?

spec/network_interface/network_interface_rh_spec.rb
```diff
@@ -0,0 +1,208 @@
...
+  describe ".new" do
+    it "loads the configuration" do
+      conf = dhcp_interface.interface_conf
+      expect(conf["NM_CONTROLLED"]).to eq("no")
+      expect(conf["DEVICE"]).to eq("eth0")
+      expect(conf["BOOTPROTO"]).to eq("dhcp")
+      expect(conf["UUID"]).to eq("3a48a5b5-b80b-4712-82f7-e517e4088999")
+      expect(conf["ONBOOT"]).to eq("yes")
+      expect(conf["TYPE"]).to eq("Ethernet")
+      expect(conf["NAME"]).to eq('"System eth0"')
```
Fryguy: For readability can you align the `eq`s

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,138 @@
+require 'ipaddr'
+require 'pathname'
+
+module LinuxAdmin
+  class NetworkInterfaceRH < NetworkInterface
+    IFACE_DIR = "/etc/sysconfig/network-scripts"
+
+    # @return [Hash<String, String>] Key value mappings in the interface file
+    attr_reader :interface_conf
```
Fryguy: Should this be private?  I know you're using it for tests, but I'm not sure it makes sense in the public interface.  Even so, can you spell out `conf` to `config`.  Perhaps there's a better word than `conf`?

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,138 @@
...
+
+    # @param interface [String] Name of the network interface to manage
+    def initialize(interface)
+      super
+      @iface_file = Pathname.new(IFACE_DIR).join("ifcfg-#{@interface}")
```
Fryguy: If you're going to name it `interface_conf`, then let's name this `@interface_file` to keep it all consistent.

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,138 @@
...
+      parse_conf
+    end
+
+    # Parses the interface configuration file into the @interface_conf hash
+    def parse_conf
```
Fryguy: Should this be private?

Fryguy:
Awesome stuff here!

lib/linux_admin/network_interface.rb
carbonin: It prevents a warning

lib/linux_admin/network_interface.rb
carbonin: How old?  I think we have travis running for 1.9.3.

lib/linux_admin/network_interface/network_interface_rh.rb
carbonin: In 1.9.3 it is an `ArgumentError`, but newer versions use `IPAddr::InvalidAddressError`, but using that blows up with "undefined constant" in 1.9.3

lib/linux_admin/network_interface/network_interface_rh.rb
carbonin: Typo, this will work for both, I used to have two different parts to handle the different types and forgot to change the message/docs

lib/linux_admin/network_interface.rb
Fryguy: OK, can't remember when that was introduce but it might be 2.0 only.

lib/linux_admin/network_interface/network_interface_rh.rb
Fryguy: Gotcha.

lib/linux_admin/network_interface.rb
carbonin: In this case it really isn't the super class new that we are looking for right?  This would be if someone called `.new` with one of the distro specific classes, then we would just call that class' `.new` method directly (which here is called `orig_new`).

Actually after thinking about this, maybe there is a better way to do this because the call stack here in the Red Hat case would look like:
NetworkInterfaceRH.new...
NetworkInterface.new line 36
NetworkInterface.new line 34
right?

So shouldn't this be:

```ruby
def self.new(*args)
  if self == LinuxAdmin::NetworkInterface
    dist_class.orig_new(*args)
  else
    orig_new(*args)
  end
end
```

lib/linux_admin/network_interface/network_interface_rh.rb
carbonin: I wanted to leave this public in case something else changed the file and there was still a reference to the object out there.

lib/linux_admin/network_interface.rb
carbonin: Removing it gives you `(Object) interface` in the documentation.

lib/linux_admin/network_interface.rb
carbonin: I'm fine with any other than just mask.  I think it is pretty intuitive either way, the terms netmask and subnet mask are used pretty interchangeably.

lib/linux_admin/network_interface.rb
carbonin: It would definitely fail if they initialized with an interface name that doesn't exist, but we are not really giving good feedback for that situation.  Maybe raising there would be better anyway?  Would we want to implement our own exception class in that case and see if we can figure out if that is what went wrong?  After giving this another look I do want to change something about it because right we are not giving any feedback at all.

lib/linux_admin/network_interface/network_interface_rh.rb
carbonin: I could be private, but I don't see an issue with giving someone the chance to see what is actually in the file, this is just exactly the key=>value pairs that will get written so it will make sense to someone who knows what the syntax is for the interface configuration.

I think `:interface_config` is good.  I don't think there is another word to use.

lib/linux_admin/network_interface/network_interface_rh.rb
carbonin: Looks like yardoc takes care of it, this is what is shows:
```
 search (Array<String>) (defaults to: nil) — list of search domains
```

Is that good enough or do you also want me to put a note about it being optional in there?

lib/linux_admin/network_interface/network_interface_rh.rb
bdunne: @carbonin Does `IPAddr::InvalidAddressError` inherit from `ArgumentError`?
If not, I think this should at least be `rescue ArgumentError, IPAddr::InvalidAddressError` to prevent rescuing unwanted errors

carbonin: Oh it does inherit from `ArgumentError`, I'll change that.

lib/linux_admin/network_interface.rb
carbonin: Yeah, it's definitely not in 1.9.3

lib/linux_admin/network_interface/network_interface_rh.rb
carbonin: Done

lib/linux_admin/network_interface.rb
```diff
@@ -0,0 +1,182 @@
...
+      parse_ip6(ip_output, :global)
+      parse_ip6(ip_output, :link)
+
+      @network_conf[:mac] = parse_ip_output(ip_output, %r{link/ether}, 1)
+
+      ip_route_res = run!(cmd("ip"), :params => ["route"])
+      @network_conf[:gateway] = parse_ip_output(ip_route_res.output, /^default/, 2) if ip_route_res.success?
+      true
+    rescue AwesomeSpawn::CommandResultError => e
+      raise NetworkInterfaceError.new(e.result, e.message)
```
bdunne: This should be message, result

lib/linux_admin/network_interface.rb
```diff
@@ -0,0 +1,182 @@
...
+    end
+
+    # Runs the command `ip addr show <interface>`
+    #
+    # @return [String] The command output
+    # @raise [NetworkInterfaceError] if the command fails
+    def ip_show
+      run!(cmd("ip"), :params => ["addr", "show", @interface]).output
+    rescue AwesomeSpawn::CommandResultError => e
+      raise NetworkInterfaceError.new(e.result, e.message)
```
bdunne: Same, message, result

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,140 @@
...
+      @interface_config["DNS1"] = servers[0]
+      @interface_config["DNS2"] = servers[1] if servers[1]
+    end
+
+    # Sets the search domain list for this network interface
+    #
+    # @param domains [Array<String>] the list of search domains
+    def search_order=(*domains)
+      domains.flatten!
+      @interface_config["DOMAIN"] = "\"#{domains.join(' ')}\""
```
bdunne: rather than modifying the args, you can
```ruby
@interface_config["DOMAIN"] = "\"#{domains.flatten.join(' ')}\""
```

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -0,0 +1,140 @@
...
+    def netmask=(mask)
+      validate_ip(mask)
+      @interface_config["NETMASK"] = mask
+    end
+
+    # Sets one or both DNS servers for this network interface
+    #
+    # @param servers [Array<String>] The DNS servers
+    def dns=(*servers)
+      servers.flatten!
```
bdunne: Maybe
```ruby
server1, server2 = servers.flatten
```

bdunne:
:+1: LGTM, will merge when green.

PR #134, opened 2015-10-16
Added method to update the IP address for a particular hostname

carbonin:
@bdunne Please review
https://trello.com/c/rWKh4KQs

### DIFF
--- a/lib/linux_admin/hosts.rb
+++ b/lib/linux_admin/hosts.rb
@@ -38,6 +38,20 @@ module LinuxAdmin
       end
     end
 
+    # Updates the IP address associated with the entry specified by the hostname
+    # or inserts a new row in the file if there is no entry for that name
+    #
+    # @param hostname [String] the hostname used to find the record to update
+    # @param ip [String] the new address to use for the record
+    def update_ip_for_hostname(host, ip)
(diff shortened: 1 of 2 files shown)


lib/linux_admin/hosts.rb
```diff
@@ -38,6 +38,20 @@ def update_entry(address, hostname, comment = nil)
...
     end
 
+    # Updates the IP address associated with the entry specified by the hostname
+    # or inserts a new row in the file if there is no entry for that name
+    #
+    # @param hostname [String] the hostname used to find the record to update
+    # @param ip [String] the new address to use for the record
+    def update_ip_for_hostname(host, ip)
+      line_num = @parsed_file.find_path(host).first
+      if line_num.nil?
```
bdunne: Prefer the positive case `if line_num`

carbonin:
Determined that this change is not necessary after discussion with @bdunne

bdunne:
The assumed use case was:
When changing IP address, update the hosts file by
- finding the old IP - hostname line
- replacing the old IP with the new IP
- save file

Setting IP should have nothing to do with the hosts file.
Setting the hostname on the machine should update the hosts file by:
- finding the line for 127.0.0.1
- removing the old hostname
- adding the new hostname

PR #137, opened 2015-10-21
Don't extend self to set up the logger

bdunne:


bdunne:
@Fryguy Please review

cc @carbonin

PR #140, opened 2015-10-26
Allow NetworkInterface creation without a config file

### LINKED ISSUE #139: LinuxAdmin::NetworkInterface.new("name") assumes there a conf file and fails
LinuxAdmin::NetworkInterface.new("interface_name") assumes there is a conf file in 
/etc/sysconfig/network-scripts, but this is not always the case.
So it fails with 

Errno::ENOENT: No such file or directory @ rb_sysopen - /etc/sysconfig/network-scripts/ifcfg-eth2

I'm running rhel 6.5, and I have an interface named eth2, and it doesn't have a conf file for it (but the network works ok)

carbonin:
This will allow a user on a Red Hat system to use the "getter" methods defined in `network_interface.rb` even if they don't have a config file corresponding to the given interface name.

To do this we instantiate a `NetworkInterfaceGeneric` object if the file doesn't exist rather than a `NetworkInterfaceRH` instance.

@abonas @bdunne @Fryguy 

Fixes #139

### DIFF
--- a/lib/linux_admin/exceptions.rb
+++ b/lib/linux_admin/exceptions.rb
@@ -6,4 +6,6 @@ module LinuxAdmin
   end
 
   class NetworkInterfaceError < AwesomeSpawn::CommandResultError; end
+
+  class MissingConfigurationFileError < StandardError; end
 end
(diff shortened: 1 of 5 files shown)


bdunne:
:+1: LGTM  @Fryguy Any concerns?

PR #144, opened 2015-12-01
Added a method to set the canonical hostname in `/etc/hosts`

### LINKED ISSUE #143: `Hosts` class should distinguish between fqdn and aliases
The `/etc/hosts` man page (http://man7.org/linux/man-pages/man5/hosts.5.html) describes a difference between the "canonical_hostname" (fqdn) and the aliases for that hostname.

The `LinuxAdmin::Hosts` class should expose that difference in its API.  Maybe a method to specifically set the first name in the list and then a separate one that appends to the aliases?

This difference is reflected in the behaviour of some external tools such as cloud-init and freeipa.
These tools will retrieve the hostname (typically via the `hostname` command or from `/etc/hostname`) then consult `/etc/hosts` to find the fqdn by taking the name in the "canonical_hostname" place in the line where the found hostname is an alias.

This can cause problems when the application acts differently based on whether the fqdn is "localhost" or not.

carbonin:
Moved the old `#update_entry` to the private method `#add_name` and aliased `#update_entry` for backwards compatibility.

@bdunne @abellotti 

Fixes #143
https://bugzilla.redhat.com/show_bug.cgi?id=1286830

### DIFF
--- a/lib/linux_admin/hosts.rb
+++ b/lib/linux_admin/hosts.rb
@@ -22,20 +22,14 @@ module LinuxAdmin
       File.write(@filename, @raw_lines.join("\n"))
     end
 
-    def update_entry(address, hostname, comment = nil)
-      # Delete entries for this hostname first
-      @parsed_file.each {|i| i[:hosts].to_a.delete(hostname)}
+    def add_alias(address, hostname, comment = nil)
+      add_name(address, hostname, false, comment)
+    end
 
-      # Add entry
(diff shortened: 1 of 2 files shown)


gtanzillo:
:+1: LGTM

carbonin:
What do you think about this @bdunne ?

abellotti:
LGTM !!

PR #145, opened 2015-12-01
Added a new line at the end of /etc/hosts

carbonin:
Previously there was no new line character at the end of the file after we save it.

This would cause a problem if someone tried to append a line to the file in some way other than using this class.

@bdunne @abellotti

### DIFF
--- a/lib/linux_admin/hosts.rb
+++ b/lib/linux_admin/hosts.rb
@@ -19,7 +19,7 @@ module LinuxAdmin
     def save
       cleanup_empty
       @raw_lines = assemble_lines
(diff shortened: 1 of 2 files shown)


abellotti:
LGTM!!  :+1:

carbonin:
@abellotti Was this causing a specific issue for you or just something you noticed while testing other hostname-related stuff?

abellotti:
noticed it while testing IPA with appliance_console.  I usually add the ipa server entry at the end of the host file and noticed its NL was gone after setting the hostname via appliance_console.

gtanzillo:
:+1: LGTM

PR #146, opened 2015-12-01
Add chrony conf class

carbonin:
Added a class for managing the chronyd configuration file.

Currently only includes methods to add servers and remove servers.

### DIFF
--- a/lib/linux_admin/chrony.rb
+++ b/lib/linux_admin/chrony.rb
@@ -1,0 +1,21 @@
+module LinuxAdmin
+  class Chrony
+    def initialize(conf = "/etc/chrony.conf")
+      raise MissingConfigurationFileError, "#{conf} does not exist" unless File.exist?(conf)
+      @conf = conf
+    end
+
+    def clear_servers
+      data = File.read(@conf)
+      data.gsub!(/^server\s+.+\n/, "")
+      File.write(@conf, data)
+    end
+
+    def add_servers(*servers)
+      data = File.read(@conf)
+      data << "\n" unless data.end_with?("\n")
+      servers.each { |s| data << "server #{s}\n" }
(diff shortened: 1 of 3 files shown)


lib/linux_admin/chrony.rb
```diff
@@ -0,0 +1,21 @@
+module LinuxAdmin
+  class Chrony
+    def initialize(conf = "/etc/chrony.conf")
+      raise MissingConfigurationFileError, "#{conf} does not exist" unless File.exist?(conf)
+      @conf = conf
+    end
+
+    def clear_servers
+      data = File.read(@conf)
+      data.gsub!(/server\s+.+\n/, "")
```
jrafanie: Is it possible to have comments or other things in this file with server in it?
Should we check that the beginning of the line starts with server?

lib/linux_admin/chrony.rb
```diff
@@ -0,0 +1,21 @@
...
+      File.write(@conf, data)
+    end
+
+    def add_servers(*servers)
+      data = File.read(@conf)
+      data << "\n" unless data.end_with?("\n")
```
jrafanie: Can we instead do `data << "\nserver #{s}"` below?

carbonin: I couldn't decide what the best way to deal with this was.
I wanted to deal with the case where the file didn't end with a newline, but also make sure the file did end with a newline when I was finished with it.

jrafanie: Ok, I'm good with that.

jrafanie:
@carbonin Can you coordinate a new release with this change so you can update your other PR to require that new version of linux_admin?

carbonin:
@jrafanie Will do

PR #147, opened 2015-12-08
Add Hardware#total_cores based on MiqSystem.num_cpus

bdunne:
https://github.com/ManageIQ/manageiq/blob/master/gems/pending/util/miq-system.rb#L48:L62


bdunne:
@jrafanie @Fryguy Please review.

We saw the incorrect core count yesterday in our debugging.

lib/linux_admin/hardware.rb
```diff
@@ -0,0 +1,7 @@
+module LinuxAdmin
+  class Hardware
+    def num_cpus
```
jrafanie: Maybe `total_cores`

bdunne:
@jrafanie Updated

jrafanie:
Looks good.

PR #148, opened 2015-12-11
Made Common its own class rather than a module

### LINKED ISSUE #19: Change subclassing of LinuxAdmin to composition by including LinuxAdmin::Common
If all we need is the behavior from LinuxAdmin::Common, we should just include the module into our classes instead of using inheritance.

carbonin:
Previously nearly all classes were mixing in LinuxAdmin::Common
to get access to what really are just utility methods.

Not sure when a good time to make this sort of change is, but I'm giving it a shot.
@bdunne @Fryguy @jrafanie 

Fixes #19

### DIFF
--- a/lib/linux_admin/common.rb
+++ b/lib/linux_admin/common.rb
@@ -6,20 +6,20 @@ module LinuxAdmin
 
     BIN_DIRS = %w(/bin /sbin /usr/bin /usr/sbin /usr/local/bin /usr/local/sbin)
 
-    def cmd(name)
+    def self.cmd(name)
       BIN_DIRS.collect { |dir| "#{dir}/#{name}" }.detect { |cmd| File.exist?(cmd) }
     end
 
-    def cmd?(name)
+    def self.cmd?(name)
       !cmd(name).nil?
     end
 
-    def run(cmd, options = {})
+    def self.run(cmd, options = {})
       AwesomeSpawn.logger ||= logger
       AwesomeSpawn.run(cmd, options)
     end
 
-    def run!(cmd, options = {})
+    def self.run!(cmd, options = {})
       AwesomeSpawn.logger ||= logger
       AwesomeSpawn.run!(cmd, options)
     end
--- a/spec/physical_volume_spec.rb
+++ b/spec/physical_volume_spec.rb
@@ -20,8 +20,8 @@ eos
     it "uses vgextend" do
       vg = LinuxAdmin::VolumeGroup.new :name => 'vg'
       pv = described_class.new :device_name => '/dev/hda'
-      expect(pv).to receive(:run!).
-         with(pv.cmd(:vgextend),
+      expect(LinuxAdmin::Common).to receive(:run!)
+        .with(LinuxAdmin::Common.cmd(:vgextend),
               :params => ['vg', '/dev/hda'])
       pv.attach_to(vg)
     end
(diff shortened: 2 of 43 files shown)


jrafanie:
This makes sense to me.  :+1:

spec/physical_volume_spec.rb
```diff
@@ -52,23 +52,23 @@
...
+        .with(LinuxAdmin::Common.cmd(:pvcreate),
+              :params => {nil => '/dev/hda'})
       described_class.create disk
     end
 
     it "returns new physical volume" do
-      allow(LinuxAdmin::VolumeGroup).to receive_messages(:run! => double(:output => ""))
-      allow(described_class).to receive_messages(:run! => double(:output => ""))
+      allow(LinuxAdmin::Common).to receive_messages(:run! => double(:output => ""))
+      allow(LinuxAdmin::Common).to receive_messages(:run! => double(:output => ""))
```
bdunne: Do you need this twice?  `allow` should handle any number of calls, right?

spec/physical_volume_spec.rb
```diff
@@ -52,23 +52,23 @@
...
       pv = described_class.create disk
       expect(pv).to be_an_instance_of(described_class)
       expect(pv.device_name).to eq('/dev/hda')
     end
 
     it "adds physical volume to local registry" do
-      allow(LinuxAdmin::VolumeGroup).to receive_messages(:run! => double(:output => ""))
-      allow(described_class).to receive_messages(:run! => double(:output => ""))
+      allow(LinuxAdmin::Common).to receive_messages(:run! => double(:output => ""))
+      allow(LinuxAdmin::Common).to receive_messages(:run! => double(:output => ""))
```
bdunne: same

spec/physical_volume_spec.rb
```diff
@@ -76,17 +76,17 @@
...
+        .and_return(double(:output => @physical_volumes))
+      expect(LinuxAdmin::Common).to receive(:run!).and_return(double(:output => @groups)) # stub out call to vgdisplay
       described_class.scan
     end
 
     it "returns local physical volumes" do
-      expect(described_class).to receive(:run!).and_return(double(:output => @physical_volumes))
-      expect(LinuxAdmin::VolumeGroup).to receive(:run!).and_return(double(:output => @groups))
+      expect(LinuxAdmin::Common).to receive(:run!).and_return(double(:output => @physical_volumes))
+      expect(LinuxAdmin::Common).to receive(:run!).and_return(double(:output => @groups))
```
bdunne: You should be able to put both doubles in a single `and_return` and the first hit will get the first response and the second will get the second response.

spec/physical_volume_spec.rb
```diff
@@ -95,8 +95,8 @@
     end
 
     it "resolves volume group references" do
-      expect(described_class).to receive(:run!).and_return(double(:output => @physical_volumes))
-      expect(LinuxAdmin::VolumeGroup).to receive(:run!).and_return(double(:output => @groups))
+      expect(LinuxAdmin::Common).to receive(:run!).and_return(double(:output => @physical_volumes))
+      expect(LinuxAdmin::Common).to receive(:run!).and_return(double(:output => @groups))
```
bdunne: same

lib/linux_admin/common.rb
```diff
@@ -1,25 +1,25 @@
 require 'awesome_spawn'
 
 module LinuxAdmin
-  module Common
+  class Common
```
bdunne: Do you need to change this?  I don't see us creating an instance anywhere and all the methods are class methods.

PR #149, opened 2016-01-07
Add process memory information: PSS, RSS, USS, shared, swap, and size.

jrafanie:
PSS should give a better indication of what a forked process is really using
since it more accurately reports how shared memory affects system memory usage.

* unique set size (USS):
sum of private mappings for the process

* proportional set size (PSS):
USS + shared_mappings.sum { |m| m.size / number_of_processes_sharing_m }

* resident set size (RSS):
USS + sum of shared mappings

[See also](http://unix.stackexchange.com/questions/33381/getting-information-about-a-process-memory-usage-from-proc-pid-smaps)

Note, the results should be comparable to [smem](https://www.selenic.com/smem/).

### DIFF
--- a/lib/linux_admin/process/memory.rb
+++ b/lib/linux_admin/process/memory.rb
@@ -1,0 +1,49 @@
+module LinuxAdmin
+  module Process
+    class Memory
+      attr_reader :pss, :rss, :uss, :shared, :swap, :size
+
+      def initialize(pid = Process.pid, smaps = nil)
+        @pss    = 0
+        @rss    = 0
+        @uss    = 0
+        @shared = 0
+        @swap   = 0
+        @size   = 0
+        @smaps_data = smaps || smaps_data(pid)
+        parse_smaps
+      end
+
+      def smaps_data(pid)
+        smaps = "/proc/#{pid}/smaps"
+        raise "smaps not found: #{smaps}" unless File.exist?(smaps)
+        File.read(smaps)
+      end
+
+      private
+
+      def parse_smaps
+        @smaps_data.each_line do |line|
+          parse_smaps_line(line)
+        end
+      end
+
+      def parse_smaps_line(line)
+        case line
+        when /^Pss:\s+([\d]+)/
+          @pss += Regexp.last_match[-1].to_i
+        when /^Rss:\s+([\d]+)/
+          @rss += Regexp.last_match[-1].to_i
+        when /^Size:\s+([\d]+)/
+          @size += Regexp.last_match[-1].to_i
+        when /^Swap:\s+([\d]+)/
+          @swap += Regexp.last_match[-1].to_i
+        when /^Private_(Clean|Dirty):\s+([\d]+)/
+          @uss += Regexp.last_match[-1].to_i
+        when /^Shared_(Clean|Dirty):\s+([\d]+)/
+          @shared += Regexp.last_match[-1].to_i
+        end
+      end
+    end
+  end
+end
--- a/lib/linux_admin.rb
+++ b/lib/linux_admin.rb
@@ -33,6 +33,7 @@ require 'linux_admin/ip_address'
 require 'linux_admin/dns'
 require 'linux_admin/network_interface'
 require 'linux_admin/chrony'
+require 'linux_admin/process'
 
 module LinuxAdmin
   extend Common
--- a/lib/linux_admin/process.rb
+++ b/lib/linux_admin/process.rb
@@ -1,0 +1,1 @@
+require_relative 'process/memory'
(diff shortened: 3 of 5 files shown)


jrafanie:
@bdunne @Fryguy @matthewd Please review / throw :tomato:
I don't really like the interface so any ideas would be helpful.

I'd like to have this so we can measure the memory usage of forked processes more accurately.

lib/linux_admin/process/memory.rb
```diff
@@ -0,0 +1,49 @@
+module LinuxAdmin
+  module Process
+    class Memory
+      attr_reader :pss, :rss, :uss, :shared, :swap, :size
```
Fryguy: Can you give meaningful-named aliases to pss, rss, uss...I'll never remember them as they are and LinuxAdmin aims for clarity over these esoteric things :)

Fryguy: Also, what is `size`? Aren't they all sizes?

lib/linux_admin/process/memory.rb
```diff
@@ -0,0 +1,49 @@
+module LinuxAdmin
+  module Process
+    class Memory
```
Fryguy: Wondering if this should be part of the Process class directly?  Just thinking about how we plan to transfer a lot of the miq-process stuff over to here.  @carbonin @bdunne Thoughts on the class layout?

jrafanie: Agreed, i wasn't convinced either way. This can move.

lib/linux_admin/process/memory.rb
jrafanie: I guess it's the same as VIRT, the total addressable memory in the process.

lib/linux_admin/process/memory.rb
```diff
@@ -0,0 +1,49 @@
...
+      def parse_smaps
+        @smaps_data.each_line do |line|
+          parse_smaps_line(line)
+        end
+      end
+
+      def parse_smaps_line(line)
+        case line
+        when /^Pss:\s+([\d]+)/
+          @pss += Regexp.last_match[-1].to_i
```
Fryguy: Would prefer `.last_match[1]` as it mirrors `$1`.  Another option is to use named matches, which might be cleaner.

lib/linux_admin/process/memory.rb
```diff
@@ -0,0 +1,49 @@
+module LinuxAdmin
+  module Process
+    class Memory
+      attr_reader :pss, :rss, :uss, :shared, :swap, :size
+
+      def initialize(pid = Process.pid, smaps = nil)
```
Fryguy: I'm not sure I like the dependency injection as part of the `.new`, because it binds you to that signature for the future and would require a major bump to change it.  Say, tomorrow a different distribution uses a different file, then you have to add a new param, and the signature is weird because callers have to pass a nil, etc.  Perhaps allow smaps_data to be set, and if set, then return it otherwise look it up...something like

```ruby
def initialize(pid = Process.pid)
  @pid = pid
  ...
end

attr_writer :smaps_data

def smaps_data
  @smaps_data ||= begin
    smaps = "/proc/#{pid}/smaps"
    raise "smaps not found: #{smaps}" unless File.exist?(smaps)
    File.read(smaps)
  end
end

# Then in the specs:

subject do
  described_class.new.tap { |o| o.smaps_data = @smaps }
end
```

I realize that changes the timing of the parse call, but maybe that shouldn't be on the `.new`?

Another option is to allow initialize to take a block and allow configuration that way?

```ruby
def initialize(pid = Process.pid)
  @pid = pid
  @pss = 0
  ...
  yield self if block_given?
  parse_smaps
end

# In spec

subject do
  described_class.new { |o| o.smaps_data = @smaps }
end

Fryguy: One more thing...won't Process here resolve to the wrong module because you're already nested in a module named Process.

Fryguy:
I know I had a lot of comments, but I really like this PR :smile: 

So here's one more...the file seems to be in kilobytes, so the output values are also in kilobytes.  I'm wondering if it should return bytes instead as that is a better base unit, particularly for conversion.  Note that, technically, kB is 1000x not 1024x,  so you might have to verify what values are *really* stored here.

lib/linux_admin/process/memory.rb
Fryguy: Also, the square brackets around the `\d` are not necessary and complicate the regex.

lib/linux_admin/process/memory.rb
Fryguy: Another option is to write a `Process#memory` method that just returns one of these Process::Memory instances.

jrafanie:
Closing in favor of adding it to the more specific sys-proctable gem:  https://github.com/djberg96/sys-proctable/pull/49

PR #151, opened 2016-01-16
Mac support

kbrock:
@Fryguy @carbonin I want to run `appliance_console` on my mac.

While I made the ip and gateway methods work, it really isn't necessary. I just want to be able to run the various menu options.

/cc @jrafanie @gtanzillo FYI

### DIFF
--- a/lib/linux_admin/network_interface.rb
+++ b/lib/linux_admin/network_interface.rb
@@ -16,6 +16,8 @@ module LinuxAdmin
       @dist_class ||= begin
         if [Distros.rhel, Distros.fedora].include?(Distros.local)
           NetworkInterfaceRH
+        elsif Distros.mac == Distros.local
+          NetworkInterfaceMac
         else
           NetworkInterfaceGeneric
         end
@@ -45,7 +47,7 @@ module LinuxAdmin
     # @raise [NetworkInterfaceError] if network information cannot be retrieved
     def reload
       @network_conf = {}
-      return false unless (ip_output = ip_show)
+      return unless (ip_output = ip_show)
 
       parse_ip4(ip_output)
       parse_ip6(ip_output, :global)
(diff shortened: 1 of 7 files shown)


kbrock:
Looks like `ip` is supported on a mac with:

```bash
brew install iproute2mac
```

So I guess this is not needed - but not standard.

Would you prefer me to continue on this PR, or we'll just go the `iproute2mac` roue?

lib/linux_admin/network_interface.rb
```diff
@@ -16,6 +16,8 @@ def self.dist_class(clear_cache = false)
       @dist_class ||= begin
         if [Distros.rhel, Distros.fedora].include?(Distros.local)
           NetworkInterfaceRH
+        elsif Distros.mac == Distros.local
+          NetworkInterfaceMac
```
carbonin: Maybe this can all be:
```ruby
case(Distros.local)
when Distros.rhel, Distros.fedora
  NetworkInterfaceRH
when Distros.mac
  NetworkInterfaceMac
else
  NetworkInterfaceGeneric
end
```
I think it's nicer than the array include followed by `==`

lib/linux_admin/time_date.rb
```diff
@@ -6,10 +6,12 @@ class TimeDate
     TimeCommandError = Class.new(StandardError)
 
     def self.system_timezone_detailed
-      result = run(cmd(COMMAND), :params => ["status"])
+      result = run!(cmd(COMMAND), :params => ["status"])
       result.output.split("\n").each do |l|
         return l.split(':')[1].strip if l =~ /Time.*zone/
       end
+    rescue AwesomeSpawn::CommandResultError, AwesomeSpawn::NoSuchFileError
+      Time.now.zone
```
carbonin: This isn't the same result as what we would get from `timedatectl`.
I'm not sure if that matters, but I don't know of a good way to convert between the `America/New_York` style and `EST`

PR #152, opened 2016-01-16
remove constants from spec

kbrock:
remove constant defined in the spec

not necessarily a fan of the `let` but you really can't define constants in these specs.

/cc @carbonin @Fryguy


carbonin:
Wasn't sure why this was needed, but then I read https://github.com/rspec/rspec-core/issues/106.  It's a bit old, but if constants are still defined outside the scope of the enclosing block this is a good call. :+1:

kbrock:
@bdunne thanks

PR #153, opened 2016-01-16
add rubocop to Gemfile

kbrock:
I use bundle exec to run commands

The bot is running `rubocops` on these. So, state `rubocop` as a dependency. (at least in the local `Gemfile`)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ spec/reports
 test/tmp
 test/version_tmp
 tmp
+bin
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -2,3 +2,5 @@ source 'https://rubygems.org'
 
 # Specify your gem's dependencies in linux_admin.gemspec
 gemspec
+
+gem 'rubocop'
```
bdunne: @kbrock shouldn't this be in the gemspec as a development dependency?

kbrock: I can move.
I like these in `Gemfile` for some reason.

kbrock:
thanks @bdunne is this good now?

PR #154, opened 2016-01-18
trim down service calls and add mac support

kbrock:
reduce duplicate code for `LinuxAdmin::Service`

add support for the mac as well.

/cc @carbonin this is related to ManageIQ/manageiq#6208


kbrock:
regarding lint - is it safe to assume `require 'bundler'` will succeed?

kbrock:
Cleaning up loose ends.

PR is way stale.

PR #156, opened 2016-02-12
Add status and reload calls to service classes

carbonin:
@bdunne

### DIFF
--- a/lib/linux_admin/service/sys_v_init_service.rb
+++ b/lib/linux_admin/service/sys_v_init_service.rb
@@ -42,5 +42,14 @@ module LinuxAdmin
 
       self
     end
+
+    def reload
+      Common.run!(Common.cmd(:service), :params => [name, "reload"])
+      self
+    end
+
+    def status
+      Common.run(Common.cmd(:service), :params => [name, "status"]).output
+    end
   end
 end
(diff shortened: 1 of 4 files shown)


lib/linux_admin/service/sys_v_init_service.rb
```diff
@@ -42,5 +42,14 @@ def restart
 
       self
     end
+
+    def reload
```
bdunne: Should both of these commands live in the service class since they are exact copies?

spec/service/sys_v_init_service_spec.rb
```diff
@@ -106,4 +106,27 @@
       expect(@service.restart).to eq(@service)
     end
   end
+
+  describe "#reload" do
```
bdunne: I feel like you can set both expectations in one test

lib/linux_admin/service/sys_v_init_service.rb
carbonin: They are close, but not exact copies. Actually they have the same differences as all the other methods.

spec/service/sys_v_init_service_spec.rb
carbonin: Yup absolutely

lib/linux_admin/service/sys_v_init_service.rb
bdunne: Ah, I missed the `service` / `systemctl` difference.

PR #157, opened 2016-02-18
Added iburst parameter while setting  ntp servers

yrudman:
https://bugzilla.redhat.com/show_bug.cgi?id=1308606

### DIFF
--- a/lib/linux_admin/chrony.rb
+++ b/lib/linux_admin/chrony.rb
@@ -14,7 +14,7 @@ module LinuxAdmin
     def add_servers(*servers)
       data = File.read(@conf)
       data << "\n" unless data.end_with?("\n")
-      servers.each { |s| data << "server #{s}\n" }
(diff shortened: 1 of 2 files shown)


chessbyte:
@jrafanie please review

jrafanie:
@yrudman I think we need to update this spec to confirm we're writing with the iburst value.

https://github.com/ManageIQ/linux_admin/blob/master/spec/chrony_spec.rb#L36

cc @carbonin

carbonin:
I `expect` we would have to update the specs :trollface: 
Are we not running travis here? @bdunne

jrafanie:
trying to kick start travis

yrudman:
Modified chrony_spec to check iburst parameter

carbonin:
:+1:

bdunne:
:+1:

PR #160, opened 2016-06-28
Adding ssh agent class

alongoldboim:
Ssh agent is used for proxy authentication on ssh connections.
We are using it when deploying an Ansible playbook on a host through ssh and giving it the ability to connect to the rest of the host without having the private key stored locally.


alongoldboim:
This change will require to stop support for ruby 1.9.3

lib/linux_admin/ssh_agent.rb
```diff
@@ -0,0 +1,85 @@
+require 'net/ssh'
+require 'fileutils'
+module LinuxAdmin
+  class SshAgent
+    attr_reader :pid
+    attr_reader :sock
+
+    def initialize(agent_socket, ssh_private_key)
+      FileUtils.mkdir_p(File.dirname(agent_socket))
```
carbonin: Can we make agent_socket optional? (from the man page of ssh-agent(1) "The default is $TMPDIR/ssh-XXXXXXXXXX/agent.<ppid>")

lib/linux_admin/ssh_agent.rb
```diff
@@ -0,0 +1,85 @@
...
+      agent_details = `ssh-agent -a #{agent_socket}`
```
carbonin: `initialize` should probably not start the agent. Maybe a `start` method, then we can write some to check if the agent is running using the pid. i.e. `running?`

lib/linux_admin/ssh_agent.rb
```diff
@@ -0,0 +1,85 @@
...
+      @pid          = agent_details.split("=")[2].split(" ")[0].chop
+      IO.popen({"SSH_AUTH_SOCK" => @sock, "SSH_AGENT_PID" => @pid}, ["ssh-add", "-"], :mode => 'w') do |f|
+        f.puts(ssh_private_key)
+        if $?.to_i != 0
+          raise StandardError, "Couldn't add key to agent"
+        end
+      end
+    end
+
+    def perform_commands(ip, username, commands, _additional_flags = "")
```
carbonin: Why not just remove the `additional_flags` param here? Is there some other API you are trying to adhere to?

linux_admin.gemspec
```diff
@@ -41,4 +41,5 @@ registration, updates, etc.
   spec.add_dependency "awesome_spawn",        "~> 1.3"
   spec.add_dependency "nokogiri"
   spec.add_dependency "openscap"
+  spec.add_dependency "net-ssh", "=3.2.0"
```
carbonin: Can this be less restrictive?

lib/linux_admin/ssh_agent.rb
```diff
@@ -0,0 +1,85 @@
...
+    attr_reader :sock
+
+    def initialize(agent_socket, ssh_private_key)
+      FileUtils.mkdir_p(File.dirname(agent_socket))
+      agent_details = `ssh-agent -a #{agent_socket}`
+      @sock         = agent_details.split("=")[1].split(" ")[0].chop
+      @pid          = agent_details.split("=")[2].split(" ")[0].chop
+      IO.popen({"SSH_AUTH_SOCK" => @sock, "SSH_AGENT_PID" => @pid}, ["ssh-add", "-"], :mode => 'w') do |f|
+        f.puts(ssh_private_key)
+        if $?.to_i != 0
```
carbonin: It doesn't seem like `IO.popen` sets `$?` until after the block, so this should be moved to the outside.
From http://ruby-doc.org/core-2.2.4/IO.html#method-c-popen:
>At the end of block, Ruby closes the pipe and sets $?. In this case IO.popen returns the value of the block.

lib/linux_admin/ssh_agent.rb
```diff
@@ -0,0 +1,85 @@
...
+          raise StandardError, "Couldn't add key to agent"
+        end
+      end
+    end
+
+    def perform_commands(ip, username, commands, _additional_flags = "")
+      result = nil
+      Net::SSH.start(ip, username, :paranoid => false, :forward_agent => true, :agent_socket_factory => -> { UNIXSocket.open(@sock) }) do |ssh|
```
carbonin: I would like this method better if it just yielded the object from `Net::SSH.start` that way you offload the behavior on error to the user and don't have to pass an array of commands.

lib/linux_admin/ssh_agent.rb
```diff
@@ -0,0 +1,85 @@
...
+            unreachable_hosts << host
+          end
+        end
+      end
+      [connection_success, unreachable_hosts]
+    end
+
+    private
+
+    def ssh_exec!(ssh, command)
```
carbonin: I feel like you could use http://net-ssh.github.io/net-ssh/Net/SSH/Connection/Session.html#method-i-exec instead of this. Thoughts?

lib/linux_admin/ssh_agent.rb
alongoldboim: Sure thing no problem changing.

lib/linux_admin/ssh_agent.rb
alongoldboim: So do we even need initialize anymore?

lib/linux_admin/ssh_agent.rb
alongoldboim: Removing.

lib/linux_admin/ssh_agent.rb
alongoldboim: Your right, moving the check from the block.

lib/linux_admin/ssh_agent.rb
alongoldboim: Ill give it a go and see how it goes :)

lib/linux_admin/ssh_agent.rb
alongoldboim: Not sure i got you on this one, can you elaborate?

carbonin:
Looks like this would require us to drop ruby 1.9.3 Thoughts @bdunne @Fryguy ?

lib/linux_admin/ssh_agent.rb
carbonin: I would think so. We would make the private key and (optional) socket the params and store them in instance variables for use in `#start`

lib/linux_admin/ssh_agent.rb
carbonin: So instead of forcing the user to pass a list we could do something like this which could allow them to do their own error handling rather than not knowing which commands were executed if one fails:

```ruby
def perform_commands(ip, username)
  Net::SSH.start(...) do |ssh|
    yield ssh
  end
end
```

We could also make the commands list optional and check for `block_given?` which would look something like:

```ruby
def perform_commands(ip, username, commands = [])
  Net::SSH.start(...) do |ssh|
    if block_given?
      yield ssh
    else
      commands.each do |cmd|
        ... stuff ...
      end
    end
  end
end
```

carbonin:
Also, adding in some simple YARD doc would be awesome, that way we get docs on ruby gems for free.
(see http://www.rubydoc.info/gems/linux_admin/0.16.0/LinuxAdmin/NetworkInterfaceRH#apply_static-instance_method)

linux_admin.gemspec
alongoldboim: Done.

lib/linux_admin/ssh_agent.rb
alongoldboim: Done.

lib/linux_admin/ssh_agent.rb
alongoldboim: I think i rather keep the current version, I find it more descriptive, plus i need the request_pty, which I'm not sure we can add that way.

lib/linux_admin/ssh_agent.rb
```diff
@@ -0,0 +1,113 @@
...
+    def stop
+      if process_exists?(@pid)
+        system({"SSH_AGENT_PID" => @pid}, "(ssh-agent -k) &> /dev/null")
+        if File.exist?(@sock)
+          File.delete(@sock)
+        end
+      end
+    end
+
+    def perform_commands(ip, username, commands)
```
carbonin: Can we make commands here have a default value so we don't have to provide it in the case we give a block?

lib/linux_admin/ssh_agent.rb
```diff
@@ -0,0 +1,113 @@
...
+            result = ssh_exec!(ssh, cmd)
+            result[:last_command] = cmd
+            break if result[:exit_status] != 0
+          end
+        end
+      end
+      result
+    end
+
+    def check_connection(ip, username, sub_ips, additional_flags)
```
carbonin: I'm not sure exactly what this is doing.
What is sub_ips? This seems to be assuming that we are connecting to a host then connecting to some other hosts from there? I feel like that wont be a very common use case.

carbonin:
@alongoldboim Can you add a PR description describing what exactly you are adding it and what the use case is? This would also be useful in the commit message.

spec/ssh_agent_spec.rb
```diff
@@ -0,0 +1,59 @@
+require 'byebug'
```
carbonin: This needs to be removed

lib/linux_admin/ssh_agent.rb
alongoldboim: changing.

lib/linux_admin/ssh_agent.rb
alongoldboim: removing this method, i can implement it using a block to perform_commands method.

lib/linux_admin/ssh_agent.rb
```diff
@@ -0,0 +1,98 @@
...
+      if $?.to_i != 0
+        raise StandardError, "Couldn't add key to agent"
+      end
+    end
+
+    def stop
+      if process_exists?(@pid)
+        system({"SSH_AGENT_PID" => @pid}, "(ssh-agent -k) &> /dev/null")
+        if File.exist?(@sock)
+          File.delete(@sock)
```
bdunne: Shouldn't we clean up the file whether the process exists or not?

alongoldboim: Fixed.

alongoldboim:
@simon3z @bdunne @Fryguy @carbonin I need this pr for the deployment feature, we will need to do another version release and bump it up in manageiq, getting this pr in will mean we have to stop supporting ruby 1.9.3 are you all good with it?
Are there more comments or we can push this forward? thanks :)

alongoldboim:
@bdunne could you write your thoughts here?

lib/linux_admin/ssh_agent.rb
```diff
@@ -0,0 +1,94 @@
+require 'net/ssh'
+require 'fileutils'
+module LinuxAdmin
+  class SshAgent
+    attr_reader :pid
+    attr_reader :sock
```
Fryguy: Please use the full word `socket` over `sock`.  No need for the abbreviation and it only leads to cognitive dissonance.

lib/linux_admin/ssh_agent.rb
```diff
@@ -0,0 +1,94 @@
...
+
+    def initialize(ssh_private_key, options = {})
+      @sock        = options[:agent_socket] if options[:agent_socket]
+      @private_key = ssh_private_key
+    end
+
+    def start
+      if @sock
```
Fryguy: Prefer using the accessor over the instance variable here and below.

lib/linux_admin/ssh_agent.rb
```diff
@@ -0,0 +1,94 @@
...
+        FileUtils.mkdir_p(File.dirname(@sock))
+        agent_details = `ssh-agent -a #{@sock}`
+      else
+        agent_details = `ssh-agent`
+        @sock         = agent_details.split("=")[1].split(" ")[0].chop
+      end
+      @pid = agent_details.split("=")[2].split(" ")[0].chop
```
Fryguy: Duplicate parsing code and its difficult readability makes me think this should be moved to a private method for readability.

lib/linux_admin/ssh_agent.rb
```diff
@@ -0,0 +1,94 @@
...
+        raise StandardError, "Couldn't add key to agent"
+      end
+    end
+
+    def stop
+      system({"SSH_AGENT_PID" => @pid}, "(ssh-agent -k) &> /dev/null") if process_exists?(@pid)
+      File.delete(@sock) if File.exist?(@sock)
+    end
+
+    def perform_commands(ip, username, commands = [])
```
Fryguy: This is really about using Net::SSH versus using SSH agent, so not sure it belongs in this class.  I'm wondering if this should be in a LinuxAdmin::SSH class or something.

lib/linux_admin/ssh_agent.rb
```diff
@@ -0,0 +1,94 @@
...
+          end
+        end
+      end
+      result
+    end
+
+    private
+
+    def process_exists?(process_pid)
+      Process.kill(0, process_pid) == 1
```
Fryguy: This method seems better suited for LinuxAdmin::Process, but it might be ok here.  @carbonin ?

lib/linux_admin/ssh_agent.rb
```diff
@@ -0,0 +1,94 @@
...
+        agent_details = `ssh-agent`
+        @sock         = agent_details.split("=")[1].split(" ")[0].chop
+      end
+      @pid = agent_details.split("=")[2].split(" ")[0].chop
+      IO.popen({"SSH_AUTH_SOCK" => @sock, "SSH_AGENT_PID" => @pid}, ["ssh-add", "-"], :mode => 'w') do |f|
+        f.puts(@private_key)
+      end
+      if $?.to_i != 0
+        raise StandardError, "Couldn't add key to agent"
+      end
```
Fryguy: Can you inline this?  `raise StandardError, "Couldn't add key to agent" if $?.to_i != 0`

Also, what is setting $?  Does the `IO.popen` do that?

Fryguy:
Structurally, this PR feels weird.  I would expect a class like `LinuxAdmin::SshAgent` to only be responsible for adding and removing keys to the agent, and perhaps starting/stopping the agent with a helper like `with_agent` that does the starting and stopping for you and lets you execute whatever you want in the block.   I would expect a different class like `LinuxAdmin::SSH` to be responsible for actually interacting over SSH.

Fryguy:
So, code-wise I would expect a caller to look like:

```ruby
ssh = LinuxAdmin::SSH.new(ip, username, private_key)
ssh.perform_commands(commands)
```

With perform_commands being implemented something like

```ruby
def perform_commands(commands)
  agent = LinuxAdmin::SSHAgent.new
  agent.add_private_key(private_key)
  agent.with_service do |socket|
    ssh = LinuxAdmin::SSH.new(ip, username, socket)
    ssh.perform_commands(commands)
  end
end
```

where with_service starts the service and ensures it stops.  This is rough design...I'm sure there are little edge cases you'd have to deal with like if the ssh-agent is already running before you start you don't want to stop it.

jrafanie:
> This change will require to stop support for ruby 1.9.3

@alongoldboim I didn't look at all the changes, what specifically stops support for ruby 1.9.3?  I'm not opposed, just curious.

alongoldboim:
@jrafanie We are using the ability to specify an ssh agent for net-ssh, this feature was added for "net-shh ~> 3.2.0" which needs ruby >= 2.0

jrafanie:
> @jrafanie We are using the ability to specify an ssh agent for net-ssh, this feature was added for "net-shh ~> 3.2.0" which needs ruby >= 2.0

Thanks, it wasn't clear if our code changes here required ruby 2.0+ or net-ssh.

lib/linux_admin/ssh_agent.rb
```diff
@@ -0,0 +1,64 @@
...
+      start
+      yield @socket
+      stop
+    rescue => e
+      stop
+      raise e
+    end
+
+    def stop
+      system({ 'SSH_AGENT_PID' => @pid }, '(ssh-agent -k) &> /dev/null') if process_exists?(@pid)
```
carbonin: This could also be `Common.run!(Common.cmd("ssh-agent"), :env => {"SSH_AGENT_PID" => @pid}, :params => [:k]) if process_exists?(@pid)`

That would allow the logger to be used, dynamically find the absolute path to the `ssh-agent` binary, and still capture the output.

lib/linux_admin/ssh_agent.rb
carbonin: If that class existed I would say yes, but I don't think it's necessary to create it in this PR.

carbonin:
In response to @Fryguy's comment (https://github.com/ManageIQ/linux_admin/pull/160#issuecomment-230587848), I feel like that ties usage of `LinuxAdmin::SSH` too closely with `ssh-agent`. I think it's a bit overkill to set up the agent for simpler tasks.

Maybe it would be better to pass an initialized `LinuxAdmin::SSHAgent` as an optional parameter to `LinuxAdmin::SSH.new` and only use the agent if it was passed. Then we could also use `SSH` with a password and not have to pass the private key to the SSH class.

Thoughts?

lib/linux_admin/ssh.rb
```diff
@@ -0,0 +1,102 @@
...
+    end
+
+    def perform_commands(commands = [], options = {})
+      result = nil
+      socket = nil
+      if options[:agent_enabled]
+        @agent = LinuxAdmin::SSHAgent.new(@private_key, :agent_socket => options[:agent_socket])
+        @agent.with_service do |socket|
+          result = if block_given?
+                     execute_commands(commands, :socket => socket, :stdin => options[:stdin], &Proc.new)
```
carbonin: Are you trying to pass the given block to `execute_commands` here?

If so, I think you would need to add `&block` to the params list of `perform_commands` and use `&block` where you have `&Proc.new` now.

alongoldboim: &Proc.new is getting the block that is passed to 'preform_commands' and passing next, i can also do it your way if you prefer, but this is working fine also in the current state.

lib/linux_admin/ssh_agent.rb
alongoldboim: WIll change, thanks!

lib/linux_admin/ssh.rb
carbonin: Ah okay, didn't know about this form. It's fine as is then.

alongoldboim:
@carbonin @bdunne besides the rubocop and the tests (need to refactor them) the rest looks ok?
I can drop the ssh_agent class if you don't see a need for it, I can do with out it.

lib/linux_admin/ssh_agent.rb
alongoldboim: That class doesn't exists currently, i think we can leave it here and if we want to build it later we can do it and refactor this class.

lib/linux_admin/ssh.rb
```diff
@@ -0,0 +1,96 @@
...
+    attr_reader :private_key
+    attr_reader :agent
+
+    def initialize(ip, username, private_key)
+      @ip = ip
+      @private_key = private_key
+      @username = username
+    end
+
+    def perform_commands(commands = [], options = {})
```
bdunne: It would be nice to see the `&block` here

lib/linux_admin/ssh.rb
```diff
@@ -0,0 +1,96 @@
...
+      result = nil
+      if options[:agent_enabled]
+        LinuxAdmin::SSHAgent.with_service(@private_key, options[:agent_socket]) do |socket|
```
bdunne: Why do you need the agent here?

bdunne: I would expect to call this on an `SSHAgent` instance.  i.e. 
```ruby
if ssh_agent
  ssh_agent.with_service do
    ...
  end
end
```

lib/linux_admin/ssh.rb
```diff
@@ -0,0 +1,96 @@
...
+    attr_reader :agent
+
+    def initialize(ip, username, private_key)
+      @ip = ip
+      @private_key = private_key
+      @username = username
+    end
+
+    def perform_commands(commands = [], options = {})
+      result = nil
```
bdunne: I don't think you need this result variable in the method.  Nothing happens after running the commands.

lib/linux_admin/ssh_agent.rb
```diff
@@ -0,0 +1,56 @@
+require 'net/ssh'
+require 'fileutils'
+module LinuxAdmin
+  class SSHAgent
```
bdunne: Should this be `LinuxAdmin::SSH::Agent`?

lib/linux_admin/ssh.rb
```diff
@@ -0,0 +1,96 @@
...
+      args = {}
+      if options[:socket]
+        args.merge!(:paranoid                   => false,
+                    :forward_agent              => true,
+                    :number_of_password_prompts => 0,
+                    :agent_socket_factory       => -> { UNIXSocket.open(options[:socket]) })
+      else
+        args[:keys] = [@private_key]
+      end
+      Net::SSH.start(@ip, @username, args) do |ssh|
```
bdunne: I think you need to `require 'net/ssh'` here

lib/linux_admin/ssh_agent.rb
```diff
@@ -0,0 +1,56 @@
...
+    attr_accessor :pid
+    attr_reader :socket
+
+    def self.with_service(private_key, socket = nil)
+      pid, socket = start(private_key, socket)
+      yield socket
+      stop(pid, socket)
```
bdunne: Should this move to `ensure`?

lib/linux_admin/ssh.rb
alongoldboim: fixed.

lib/linux_admin/ssh.rb
alongoldboim: it should return the stdout stderr...

lib/linux_admin/ssh_agent.rb
alongoldboim: right moving.

lib/linux_admin/ssh.rb
alongoldboim: added.

bdunne: I think I'd prefer to see the `require net/ssh` right before the call rather than at the top of the file.  Otherwise everyone that requires `linux_admin` also gets `net/ssh` loaded even if they don't plan to use it.

In a large app like ManageIQ/manageiq we try to delay load as much as possible to reduce the memory footprint.  i.e. There's no reason for some worker to load `net/ssh`.

alongoldboim: Your right, moving it.

bdunne:
@carbonin I agree with the idea of not requiring the agent.  There are simple use cases where you may want to ssh to a box, run a command and that's it.

alongoldboim:
@bdunne You don't have to use the agent, its optional.
```
ssh = LinuxAdmin::SSH.new(ip, username, private_key)
result = ssh.perform_commands(["ls"])
```

lib/linux_admin/ssh_agent.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    def start
+      if @socket
+        FileUtils.mkdir_p(File.dirname(@socket))
+        agent_details = `ssh-agent -a #{@socket}`
+      else
+        agent_details = `ssh-agent`
+        @socket = parse_ssh_agent_socket agent_details
+      end
+      @pid = parse_ssh_agent_pid agent_details
```
bdunne: Please use `method(params)` syntax for params.

bdunne:
To build on Jasons previous comment, the list of commands also seems strange to @carbonin and me.
```ruby
def perform_commands(commands)
  agent = LinuxAdmin::SSHAgent.new
  agent.add_private_key(private_key)
  agent.with_service do |socket|
    ssh = LinuxAdmin::SSH.new(ip, username, socket)
    ssh.execute(command_1)
    ssh.execute(command_2)
    ssh.execute(command_3)
  end
end
```

This might also help to clean up the result and error handling.

alongoldboim:
OK, so this is pretty much what I'm doing but that code piece has some problems(unless im getting something wrong): 
1) 'perform_commands' is an instance method, why calling 'LinuxAdmin::SSH.new(ip, username, socket)'  ? examples how its used from the caller:
with block:
 https://github.com/ManageIQ/manageiq/pull/9230/files#diff-6e045b8ce93dfa6ccec7115f7cd1cba0R67
without block: 
https://github.com/ManageIQ/manageiq/pull/9230/files#diff-6e045b8ce93dfa6ccec7115f7cd1cba0R61
2) execute_commands is equal to ssh.execute, so whats the difference?

lib/linux_admin/ssh.rb
```diff
@@ -0,0 +1,94 @@
+module LinuxAdmin
+  class SSH
+    attr_reader :ip
+    attr_reader :username
+    attr_reader :private_key
+    attr_reader :agent
+
+    def initialize(ip, username, private_key)
```
carbonin: The private key is only needed when using ssh agent. This is one of the reasons why I think we should be passing in the socket here like @bdunne suggested here https://github.com/ManageIQ/linux_admin/pull/160#issuecomment-232712525 from a previously initialized agent which the caller has set up explicitly.

Also it would also be good to have an optional password argument here in case the user does not have the auth keys set up.

lib/linux_admin/ssh.rb
```diff
@@ -0,0 +1,94 @@
...
+      @ip = ip
+      @private_key = private_key
+      @username = username
+    end
+
+    def perform_commands(commands = [], options = {})
+      require 'net/ssh'
+      if options[:agent_enabled]
```
carbonin: This could also be improved if the socket was passed in the initializer. That way we don't have to reference the SSHAgent class from here at all.

lib/linux_admin/ssh.rb
```diff
@@ -0,0 +1,94 @@
...
+      elsif block_given?
+        execute_commands(commands, :stdin => options[:stdin], &Proc.new)
+      else
+        execute_commands(commands, :stdin => options[:stdin])
+      end
+    end
+
+    private
+
+    def execute_commands(commands = [], options = {})
```
carbonin: I think maybe individual parameters would be better than an options hash here, even if they have to be optional. This is a private method so we don't have to worry about keeping a specified API and parameters help to show what can be done with the method much better than an options hash.

Also, why is `commands` optional? Why would someone call this without specifying commands to be run?

lib/linux_admin/ssh_agent.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    def start
+      if @socket
+        FileUtils.mkdir_p(File.dirname(@socket))
+        agent_details = `ssh-agent -a #{@socket}`
+      else
+        agent_details = `ssh-agent`
+        @socket = parse_ssh_agent_socket(agent_details)
+      end
+      @pid = parse_ssh_agent_pid agent_details
```
carbonin: Can we add parens around this method call to be consistent?

carbonin:
I'm okay with this. :+1: 
It looks like the bot complaining about the `puts` is actually a `File#puts` so that can likely be ignored.

I'll leave the last review to @bdunne or @Fryguy, they have the green button after all :smile:

lib/linux_admin/ssh_agent.rb
alongoldboim: fixed :+1:

alongoldboim:
@simon3z @bdunne any chance to do another release so i could bump it on manageiq?

PR #161, opened 2016-07-07
Drop activesupport dependency

bdunne:
Drop support for Ruby 1.9.3 and test against 2.3.1
Rails 5 isn't compatible with older rubies.
Rather than dropping support for older rubies just for #blank?, we can define #blank? if it isn't already defined.

### DIFF
--- a/lib/linux_admin/disk.rb
+++ b/lib/linux_admin/disk.rb
@@ -13,21 +13,25 @@ module LinuxAdmin
     def str_to_bytes(val, unit)
       case unit
       when 'K' then
-        val.to_f.kilobytes
+        val.to_f * 1_024 # 1.kilobytes
       when 'M' then
-        val.to_f.megabytes
+        val.to_f * 1_048_576 # 1.megabyte
       when 'G' then
-        val.to_f.gigabytes
+        val.to_f * 1_073_741_824 # 1.gigabytes
       end
     end
-    
+
     def overlapping_ranges?(ranges)
       ranges.find do |range1|
         ranges.any? do |range2|
           range1 != range2 &&
-          range1.overlaps?(range2)
+            ranges_overlap?(range1, range2)
         end
       end
+    end
+
+    def ranges_overlap?(range1, range2) # copied from activesupport Range#overlaps?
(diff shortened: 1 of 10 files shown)


bdunne:
@Fryguy Please review

bdunne:
@carbonin Please review.

lib/linux_admin/disk.rb
```diff
@@ -13,23 +13,27 @@ class Disk
...
       case unit
       when 'K' then
-        val.to_f.kilobytes
+        val.to_f * 1_024 # 1.kilobytes
       when 'M' then
-        val.to_f.megabytes
+        val.to_f * 1_048_576 # 1.megabyte
       when 'G' then
-        val.to_f.gigabytes
+        val.to_f * 1_073_741_824 # 1.gigabytes
```
carbonin: So, *technically* a kilobyte is 1000 bytes right? https://en.wikipedia.org/wiki/Kilobyte has a decent chart, but which form is being used is usually ambiguous. I'm assuming you checked out that these were what `1.kilobytes` and such actually resolved to?

Fryguy:
```ruby
[1] pry(main)> 1.gigabytes
=> 1073741824
```

lib/linux_admin/registration_system/subscription_manager.rb
```diff
@@ -34,7 +34,7 @@ def organizations(options)
       params.merge!(proxy_params(options))
 
       result = run!(cmd, :params => params)
-      parse_output(result.output).index_by {|i| i[:name]}
+      parse_output(result.output).each_with_object({}) { |i, h| h[i[:name]] = i }
```
Fryguy: Was about to suggest using `.index_by`...LOL...I would have failed that question if it came up on http://railshurts.com/quiz/

carbonin:
LGTM :+1:

lib/linux_admin/disk.rb
bdunne: Yes, I used the value of `1.gigabyte`, `1.megabyte`, `1.kilobyte`, etc.

PR #164, opened 2016-08-03
Add versioning to the nokogiri dependency

carbonin:
This should fix the Hakiri security warnings

@bdunne please review

Is there a particular reason that nokogiri wasn't versioned before?


Fryguy:
@carbonin It's open ended so we don't have to constantly bump it for new versions.  We can probably limit to just two digits in the gemspec `~> 1.6`

carbonin:
Will do

carbonin:
Looks like that does't do it for hakiri. Want to just close this then? @Fryguy

PR #165, opened 2016-08-15
Changed number_of_password_prompts to be 0 in all cases

alongoldboim:
SSH class will give a prompt instead of failing authentication without this attributes.


alongoldboim:
@bdunne can you review please?

PR #166, opened 2016-08-29
Improve error message when yum check-update returns error

isimluk:
The `yum check-update` will return 1 when the system has connectivity
issues or when the system is not registered with rhn (empty repolist).

Without the improvement, users have low chances to understand what was
the nature of an error, as they may not understand what command was
being run.

Addressing:
https://bugzilla.redhat.com/show_bug.cgi?id=1369556

@miq-bot add_label bug, core

### DIFF
--- a/lib/linux_admin/yum.rb
+++ b/lib/linux_admin/yum.rb
@@ -43,11 +43,11 @@ module LinuxAdmin
       cmd    = "yum check-update"
       params = {nil => packages} unless packages.blank?
 
-      exitstatus = Common.run(cmd, :params => params).exit_status
-      case exitstatus
+      spawn = Common.run(cmd, :params => params)
+      case spawn.exit_status
       when 0;   false
       when 100; true
-      else raise "Error: Exit Code #{exitstatus}"
+      else raise "Error: #{cmd} returns '#{spawn.exit_status}', '#{spawn.error}'"
       end
     end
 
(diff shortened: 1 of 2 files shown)


Fryguy:
@carbonin @bdunne Please review

lib/linux_admin/yum.rb
```diff
@@ -43,11 +43,11 @@ def self.updates_available?(*packages)
...
       params = {nil => packages} unless packages.blank?
 
-      exitstatus = Common.run(cmd, :params => params).exit_status
-      case exitstatus
+      spawn = Common.run(cmd, :params => params)
+      case spawn.exit_status
       when 0;   false
       when 100; true
-      else raise "Error: Exit Code #{exitstatus}"
+      else raise "Error: #{cmd} returns '#{spawn.exit_status}', '#{spawn.error}'"
```
carbonin: Maybe also print `spawn.output` here in case there is something on stdout?

isimluk: There is always something on stdout. Like the list of enabled plug-ins.

In my previous life (around 2010) I have been writing automated tests for yum. At that time the tool was pretty rigid on putting errors on stderr and user messages to stdout.

carbonin: In that case this is fine. :+1:

bdunne:
Reopening to kick Travis

PR #167, opened 2016-09-07
Rails 5 dropped support for ruby < 2.2.2, allow tests to run with this hack

bdunne:
The dependency on `more_core_extensions` pulls in `activesupport`


bdunne:
@Fryguy Please review

PR #169, opened 2016-10-03
Add a method to list the timezones in TimeDate

carbonin:
This will accurately show the timezones available for use with `TimeDate.system_timezone=`

Motivated by https://bugzilla.redhat.com/show_bug.cgi?id=1356688


carbonin:
@bdunne please review

PR #170, opened 2016-10-24
Add a method to set the hostname for both IPv6 and IPv4 loopback addr…

jvlcek:
This PR will add a new method to the calling API that can be used to set both the IPv6 and IPv4 loopback hostnames with a single  call.

### DIFF
--- a/lib/linux_admin/hosts.rb
+++ b/lib/linux_admin/hosts.rb
@@ -26,6 +26,10 @@ module LinuxAdmin
 
     alias_method :update_entry, :add_alias
 
+    def set_loopback_hostname(hostname, comment = nil)
+      ["::1", "127.0.0.1"].each { |address| add_name(address, hostname, true, comment, false) }
+    end
+
     def set_canonical_hostname(address, hostname, comment = nil)
       add_name(address, hostname, true, comment)
     end
@@ -46,9 +50,9 @@ module LinuxAdmin
 
     private
 
-    def add_name(address, hostname, fqdn, comment)
+    def add_name(address, hostname, fqdn, comment, remove_existing = true)
       # Delete entries for this hostname first
-      @parsed_file.each { |i| i[:hosts].to_a.delete(hostname) }
+      @parsed_file.each { |i| i[:hosts].to_a.delete(hostname) } if remove_existing
 
       # Add entry
       line_number = @parsed_file.find_path(address).first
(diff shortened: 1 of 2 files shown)


jvlcek:
@carbonin Please review

carbonin:
@jvlcek can you do the spec refactoring in a separate commit so the changes you made are easier to review?

jvlcek:
@carbonin I had done the spec refactoring to address RuboCop warnings in a separate commit but I squashed my commits since you had already commented `Looks good.`

carbonin:
@bdunne can you take a look when you get a chance?

jvlcek:
PR: https://github.com/ManageIQ/manageiq/pull/12171 depends on this change, which
will address BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1360928

lib/linux_admin/version.rb
```diff
@@ -1,3 +1,3 @@
 module LinuxAdmin
-  VERSION = "0.18.0"
+  VERSION = "0.19.0".freeze
```
bdunne: Don't change the version in a PR.  I will update it at release time.

lib/linux_admin/hosts.rb
```diff
@@ -26,6 +26,10 @@ def add_alias(address, hostname, comment = nil)
 
     alias_method :update_entry, :add_alias
 
+    def set_loopback_hostname(hostname, comment = nil)
+      ["::1", "127.0.0.1"].each { |address| add_name(address, hostname, true, comment, false) }
```
bdunne: Should we worry about this method adding duplicates if the hostname already exists for any of the loopback addresses?

jvlcek: @bdunne That had crossed my mind too. It turns out duplicates are removed already here:
https://github.com/jvlcek/linux_admin/blob/09f99dabc5578676432d609b6a069e3dbd685d70/lib/linux_admin/hosts.rb#L101

lib/linux_admin/version.rb
jvlcek: OK. Will do

bdunne:
@jvlcek Can you squash the revert commits?

lib/linux_admin/hosts.rb
bdunne: Perfect, thanks

PR #174, opened 2017-02-07
Add a #show method to the service classes

carbonin:
For systemd `show` prints some useful metrics about the unit and the cgroup at runtime.

This can be used to get memory used, memory/cpu limits and the main PID of the unit.
https://www.freedesktop.org/software/systemd/man/systemctl.html#show%20PATTERN...%7CJOB...

There is no equivalent call for SysV scripts so we raise `NotImplementedError` for the show method there.
https://fedoraproject.org/wiki/SysVinit_to_Systemd_Cheatsheet

@jrafanie we would have to then look at the `MemoryCurrent` key in the returned hash. I was debating creating a separate method for `memory`, but wasn't sure about how far I wanted to diverge the API of `SysVInitService` and `SystemdService`

@bdunne please review

### DIFF
--- a/lib/linux_admin/service/systemd_service.rb
+++ b/lib/linux_admin/service/systemd_service.rb
@@ -45,10 +45,28 @@ module LinuxAdmin
       Common.run(command_path, :params => ["status", name]).output
     end
 
+    def show
+      output = Common.run!(command_path, :params => ["show", name]).output
+      output.split("\n").each_with_object({}) do |line, h|
+        k, v = line.split("=", 2)
+        h[k] = cast_show_value(k, v)
+      end
+    end
+
     private
 
     def command_path
       Common.cmd(:systemctl)
+    end
+
+    def cast_show_value(key, value)
+      return value.to_i if value =~ /^\d+$/
+      return parse_exec_value(value) if key =~ /Exec(Start|Stop)/
+      value
+    end
+
+    def parse_exec_value(value)
+      value[1..-2].strip.split(" ; ").map { |s| s.split("=", 2) }.to_h
     end
   end
 end
--- a/lib/linux_admin/service/systemd_service.rb
+++ b/lib/linux_admin/service/systemd_service.rb
@@ -61,8 +61,15 @@ module LinuxAdmin
 
     def cast_show_value(key, value)
       return value.to_i if value =~ /^\d+$/
-      return parse_exec_value(value) if key =~ /Exec(Start|Stop)/
-      value
+
+      case key
+      when /^.*Timestamp$/
+        Time.parse(value)
+      when /Exec(Start|Stop)/
+        parse_exec_value(value)
+      else
+        value
+      end
     end
 
     def parse_exec_value(value)
--- a/spec/service/systemd_service_spec.rb
+++ b/spec/service/systemd_service_spec.rb
@@ -110,6 +110,7 @@ describe LinuxAdmin::SystemdService do
     it "returns a hash of runtime information" do
       output = <<-EOS
 MainPID=29189
+ExecMainStartTimestamp=Wed 2017-02-08 13:49:57 EST
 ExecStart={ path=/bin/sh ; argv[]=/bin/sh -c /bin/evmserver.sh start ; status=0/0 }
 ExecStop={ path=/bin/sh ; argv[]=/bin/sh -c /bin/evmserver.sh stop ; status=0/0 }
 ControlGroup=/system.slice/evmserverd.service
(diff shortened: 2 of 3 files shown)


lib/linux_admin/service/systemd_service.rb
```diff
@@ -44,12 +37,23 @@ def restart
...
     end
 
     def status
-      Common.run(Common.cmd(:systemctl), :params => ["status", name]).output
+      Common.run(command_path, :params => ["status", name]).output
+    end
+
+    def show
+      output = Common.run!(command_path, :params => ["show", name]).output
+      output.split("\n").map { |line| line.split("=", 2) }.to_h
```
bdunne: Generally I would prefer `collect` over `map` (I thought we had that in the Rubocop config), but in this case I would recommend `each_with_object({})`.

spec/service/systemd_service_spec.rb
```diff
@@ -116,4 +116,29 @@
...
+
+      hash = {
+        "MainPID"       => "29189",
+        "ExecStart"     => "{ path=/bin/sh ; argv[]=/bin/sh -c /bin/evmserver.sh start ; status=0/0 }",
+        "ExecStop"      => "{ path=/bin/sh ; argv[]=/bin/sh -c /bin/evmserver.sh stop ; status=0/0 }",
+        "ControlGroup"  => "/system.slice/evmserverd.service",
+        "MemoryCurrent" => "2865373184"
+      }
+      expect(LinuxAdmin::Common).to receive(:run!)
+        .with(LinuxAdmin::Common.cmd(:systemctl),
```
bdunne: Would it make sense to put the command args in another variable to improve readability by allowing the `expect` to be written on one line?

lib/linux_admin/service/systemd_service.rb
Fryguy: I'm fine with this way...I think this is more elegant than what each_with_object looks like.

spec/service/systemd_service_spec.rb
```diff
@@ -116,4 +116,29 @@
...
+ControlGroup=/system.slice/evmserverd.service
+MemoryCurrent=2865373184
+EOS
+
+      hash = {
+        "MainPID"       => "29189",
+        "ExecStart"     => "{ path=/bin/sh ; argv[]=/bin/sh -c /bin/evmserver.sh start ; status=0/0 }",
+        "ExecStop"      => "{ path=/bin/sh ; argv[]=/bin/sh -c /bin/evmserver.sh stop ; status=0/0 }",
+        "ControlGroup"  => "/system.slice/evmserverd.service",
+        "MemoryCurrent" => "2865373184"
```
Fryguy: I think we should cast whatever makes sense.

spec/service/systemd_service_spec.rb
```diff
@@ -116,4 +116,29 @@
...
+ExecStart={ path=/bin/sh ; argv[]=/bin/sh -c /bin/evmserver.sh start ; status=0/0 }
+ExecStop={ path=/bin/sh ; argv[]=/bin/sh -c /bin/evmserver.sh stop ; status=0/0 }
+ControlGroup=/system.slice/evmserverd.service
+MemoryCurrent=2865373184
+EOS
+
+      hash = {
+        "MainPID"       => "29189",
+        "ExecStart"     => "{ path=/bin/sh ; argv[]=/bin/sh -c /bin/evmserver.sh start ; status=0/0 }",
+        "ExecStop"      => "{ path=/bin/sh ; argv[]=/bin/sh -c /bin/evmserver.sh stop ; status=0/0 }",
```
Fryguy: I also think we should parse this...maybe something like

```ruby
def parse_exec_line(line)
  line[1..-2].split(" ; ").map { |s| s.split("=", 2).map(&:strip) }.to_h
end
```

lib/linux_admin/service/systemd_service.rb
Fryguy:
```ruby
output.split("\n").each_with_object({}) do |line, hash|
  k, v = line.split("=", 2)
  h[k] = v
end
```

I think that's uglier, personally

carbonin: I agree with @Fryguy, but I think I'm going to go @bdunne's route as I'll need to add some additional key/value logic to determine if I should cast or parse the value into some other form.

carbonin:
@Fryguy @bdunne I extracted the cleanup type stuff to https://github.com/ManageIQ/linux_admin/pull/175 to make this a bit more targeted.

carbonin:
Okay, I also added parsing out timestamp keys to a `Time` object.

The only things were not parsing into ruby objects are things like this:
```
RestartUSec=100ms
TimeoutStartUSec=1min 30s
TimeoutStopUSec=1min 30s
```

Not sure what kind of object or parsing we would do for such a thing.

jrafanie:
LGTM, what is the hakiri warning?  I don't see a warning.  😕

carbonin:
@jrafanie I tried to fix a while back it but Hakiri's requirement in the gemspec is a bit more strict than we want it to be (https://github.com/ManageIQ/linux_admin/pull/164). It looks like all the recently merged PRs have this failure.

jrafanie:
LGTM, hakiri warnings are for a "loose" nokogiri dependency unrelated to this PR.

PR #176, opened 2017-03-07
Network interface s/netmask6/prefix6/

isimluk:
With IPv6, the literature prefers to talk about `prefix length`. The netmask is still technically correct, but only rarely used.

We don't expect our users to read&type stuff like: `"ffff:ffff:ffff:ffff:ffff:ffff::"` we want them to say just `96` or `42`.

### DIFF
--- a/lib/linux_admin/network_interface.rb
+++ b/lib/linux_admin/network_interface.rb
@@ -80,6 +80,18 @@ module LinuxAdmin
       end
     end
 
+    # Retrieve the prefix length related to IPv6 address assigned to the interface
+    #
+    # @return [String] IPv6 prefix length
+    # @raises [ArgumentError] if the given scope is not `:global` or `:link`
+    def prefix6(scope = :global)
+      if [:global, :link].include?(scope)
+        @network_conf["prefix6_#{scope}".to_sym]
+      else
+        raise ArgumentError, "Unrecognized address scope #{scope}"
+      end
+    end
+
     # Retrieve the MAC address associated with the interface
     #
     # @return [String] the MAC address
@@ -92,20 +104,6 @@ module LinuxAdmin
     # @return [String] IPv4 netmask
     def netmask
       @network_conf[:mask]
-    end
-
-    # Retrieve the IPv6 sub-net mask assigned to the interface
-    #
-    # @return [String] IPv6 netmask
-    # @raise [ArgumentError] if the given scope is not `:global` or `:link`
-    def netmask6(scope = :global)
-      if scope == :global
-        @network_conf[:mask6_global]
-      elsif scope == :link
-        @network_conf[:mask6_link]
-      else
-        raise ArgumentError, "Unrecognized address scope #{scope}"
-      end
     end
 
     # Retrieve the IPv4 default gateway associated with the interface
@@ -168,13 +166,12 @@ module LinuxAdmin
     # @param ip_output [String] The command output
     # @param scope     [Symbol] The IPv6 scope (either `:global` or `:local`)
     def parse_ip6(ip_output, scope)
-      mask_addr = IPAddr.new('ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff')
       cidr_ip = parse_ip_output(ip_output, /inet6 .* scope #{scope}/, 1)
       return unless cidr_ip
 
       parts = cidr_ip.split('/')
       @network_conf["address6_#{scope}".to_sym] = parts[0]
-      @network_conf["mask6_#{scope}".to_sym] = mask_addr.mask(parts[1]).to_s
+      @network_conf["prefix6_#{scope}".to_sym] = parts[1]
     end
   end
 end
--- a/spec/network_interface_spec.rb
+++ b/spec/network_interface_spec.rb
@@ -181,17 +181,17 @@ IP_OUT
       end
     end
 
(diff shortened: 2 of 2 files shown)


isimluk:
Nothing under ManageIQ ☂️  uses `netmask6`, so we can safely replace.

isimluk:
> I think that if netmask is still technically correct, I would rather have the interface be consistent between 
> IPv4 and IPv6 than have the naming be slightly more correct.

Nod. Although, I believe this goes deeper.

Let's take a look what others do.
 * The output of `ip` command uses prefixes exclusively for both ipv4 and ipv6.
 * The `nmcli` tool seems to use prefixes exclusivelly.
 * The syntax of NetworkManager config files requires netmask for ipv4 and prefixes in ipv6. The ``
 * The UI tools (like cockpit or NetworkManager-gnome) use netmaks for ipv4 and prefix length for ipv6.

Note that NM config, ip command and nmcli command take `*/prefix` as a essential part of ipaddress. There are no specific way to say `netmask` they just say:  IPADDR=`::1/127`. On modern Linux, the only low-level place that takes netmask seems to be `NETMASK` in NM config.

My theory is that people are used to netmask for IPv4 addresses because they remember the first time they connected their device to a network. They remember when CIDR was the new trick they learned. However, they never used to see netmask6.

So, I am inclined to allow for the slight API inconsistency. What about rewriting IPv4 API to use prefix as well?

isimluk:
> Also, even if we were to go this route, it would be good to preserve the yardoc and have a separate
> commit for the refactoring rather than a commit for removing and then re-adding a functionally
> equivalent method.

Wrote the docs and squashed the commits.

The motivation for having the commits separated was that it is not refactoring, rather removal of unused method and addition of a method that will be useful. It is not clear for outside contributor if this project has aspiration to become a library or not. If the aspiration is there, then we should keep the interface intact. And I would need to drop the first commit. :)

jrafanie:
> I think that if netmask is still technically correct, I would rather have the interface be consistent between
> IPv4 and IPv6 than have the naming be slightly more correct.
> Nod. Although, I believe this goes deeper.
> 
> Let's take a look what others do.
> 
> The output of ip command uses prefixes exclusively for both ipv4 and ipv6.
> The nmcli tool seems to use prefixes exclusivelly.
> The syntax of NetworkManager config files requires netmask for ipv4 and prefixes in ipv6. The ``
> The UI tools (like cockpit or NetworkManager-gnome) use netmaks for ipv4 and prefix length for ipv6.
> Note that NM config, ip command and nmcli command take */prefix as a essential part of ipaddress. There are no specific way to say netmask they just say: IPADDR=::1/127. On modern Linux, the only low-level place that takes netmask seems to be NETMASK in NM config.
> 
> My theory is that people are used to netmask for IPv4 addresses because they remember the first time they connected their device to a network. They remember when CIDR was the new trick they learned. However, they never used to see netmask6.
> 
> So, I am inclined to allow for the slight API inconsistency. What about rewriting IPv4 API to use prefix as well?

@isimluk @carbonin I agree with both of you.  We should be consistent between both IPv4 and v6 API but clearly the tools aren't all consistent.  Even though we're probably not using the current netmask6, technically renaming this is a breaking change and we should deprecate it before removing it.

I'm wondering if we can solve all of this by aliasing them so we have both prefix/prefix6 and netmask/netmask6.  The implementation can use `prefix/prefix6`  with the netmasks aliasing prefix.  Our client code can then choose to use either one.   `prefix` if that's the more widely used, or `netmask` if the client code is in an area that consistently uses `netmask`.

jrafanie:
> Wrote the docs and squashed the commits.
> 
> The motivation for having the commits separated was that it is not refactoring, rather removal of unused method and addition of a method that will be useful. It is not clear for outside contributor if this project has aspiration to become a library or not. If the aspiration is there, then we should keep the interface intact. And I would need to drop the first commit. :)

@isimluk to be fair, this is a trivial change so it's pretty obvious what's going on.  But, it's a good habit to make it easy for others see exactly what was done.  If the remove and add with a different name are separate commits and in a much larger pull request, it would be hard to follow what was going on.  By doing the rename in one commit and any refactorings in another, me, others, future you, can easily walk the commits one by one and know exactly what happened.  It's always possible to squash commits later if they don't need to be so small but to be honest, I can't recall ever disliking small "singular functional change" commits.  Squashing typos, fixes for prior commits, or the same functional change split among several commits is perfectly fine.

carbonin:
>I'm wondering if we can solve all of this by aliasing them so we have both prefix/prefix6 and netmask/netmask6. The implementation can use prefix/prefix6 with the netmasks aliasing prefix. Our client code can then choose to use either one. prefix if that's the more widely used, or netmask if the client code is in an area that consistently uses netmask.

I like this idea, but I'm not sure an alias will do the job. Netmask is a full address where prefix is just a number so we can't possibly have one method that returns both.

Maybe it would be better to have separate methods that interpret the same value on the object in different ways? We can keep the backend hash value as a prefix then when someone calls `netmask6` we return that prefix as an IPv6 netmask. This way we add functionality and don't break the API.

Fryguy:
:+1: for aliasing.  In particular, removing netmask6 requires a major version bump in the gem.  I don't necessarily care about the version number, but we would also have to release a minor version with a deprecation warning.

isimluk:
Ok, I'll start over.

PR #177, opened 2017-03-07
Support for showing IPv6 default gateway

isimluk:


lib/linux_admin/network_interface.rb
```diff
@@ -112,7 +114,11 @@ def netmask6(scope = :global)
     #
     # @return [String] IPv4 gateway address
     def gateway
-      @network_conf[:gateway]
+      @network_conf[:gateway4]
+    end
+
```
carbonin: Can we continue to write documentation for these methods?

lib/linux_admin/network_interface.rb
```diff
@@ -51,8 +51,10 @@ def reload
 
       @network_conf[:mac] = parse_ip_output(ip_output, %r{link/ether}, 1)
 
-      ip_route_res = Common.run!(Common.cmd("ip"), :params => ["route"])
-      @network_conf[:gateway] = parse_ip_output(ip_route_res.output, /^default/, 2) if ip_route_res.success?
+      [4, 6].each do |version|
```
carbonin: This all might be cleaner if we move the call to `run!` and the rescue into a separate method (`ip_route(version)` maybe?).
Then the assignment will just be two lines like:
```ruby
@network_conf[:gateway]  = parse_ip_output(ip_route(4), /^default/, 2)
```

That should also make this all easier to unit test.

lib/linux_admin/network_interface.rb
isimluk: Added documentation.

Previously, I was kinda reluctant to write it as the other documentation strings are not 100% accurate. (for instance, some stuff here relies on the appliance having single network_interface).

lib/linux_admin/network_interface.rb
isimluk: Done.

isimluk:
Amended based on feedback. Thanks!

lib/linux_admin/network_interface.rb
carbonin: Yeah, the single/multiple interface thing will make all of this much more difficult. :worried: I'm not sure how we deal with that currently ...

carbonin: I guess it's easier for the commands that use `ip addr`, but `ip route` doesn't take an interface name ... Is there something we can do to make that less fragile?

isimluk: Ha found it. `ip -6 route show dev #{name}` I can do it in follow-up pr.

Or we can wait until it is broken for somebody. We are still using old network device naming `eth0` and not the dynamic one. As a side effect the ethX can change in order cause a disruption.

So my answer to anyone with custom setup would be: don't use appliance_console, use what you have learned in RHCSA. :)

carbonin: Good find! I think a follow up would be good for this. It would be a general enhancement around making things work for multi-nic setups.

carbonin:
Looking good. I think we still need a spec for the `#gateway6` method.

isimluk:
Thanks! Spec added.

isimluk:
The harakiri failure seems unrelated to this pr.

PR #178, opened 2017-03-08
Introduce methods for setting up static ipv6

isimluk:
# What
 - Introduce `address6=` method for setting static IPv6 address in /etc/sysconfig/network-scripts/ifcfg
 - introduce `gateway6=` method
 - introduce `apply_static6` method

### DIFF
--- a/lib/linux_admin/network_interface/network_interface_rh.rb
+++ b/lib/linux_admin/network_interface/network_interface_rh.rb
@@ -60,6 +60,15 @@ module LinuxAdmin
       @interface_config["GATEWAY"] = address
     end
 
+    # Set the IPv6 gateway address for this interface
+    #
+    # @param address [String] IPv6 address optionally including the prefix length
+    # @raise ArgumentError if the address is not formatted properly
+    def gateway6=(address)
+      validate_ip(address)
+      @interface_config['IPV6_DEFAULTGW'] = address
+    end
+
     # Set the IPv4 sub-net mask for this interface
     #
     # @param mask [String]
--- a/lib/linux_admin/network_interface/network_interface_rh.rb
+++ b/lib/linux_admin/network_interface/network_interface_rh.rb
@@ -125,6 +125,23 @@ module LinuxAdmin
       save
     end
 
+    # Applies the given static IPv6 network configuration to the interface
+    #
+    # @param ip [String] IPv6 address
+    # @param prefix [Number] prefix length for IPv6 address
+    # @param gw [String] gateway address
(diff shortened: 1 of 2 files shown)


lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -104,6 +125,23 @@ def apply_static(ip, mask, gw, dns, search = nil)
...
+    # @param gw [String] gateway address
+    # @param dns [Array<String>] list of dns servers
+    # @param search [Array<String>] list of search domains
+    # @return [Boolean] true on success, false otherwise
+    # @raise ArgumentError if an IP is not formatted properly or interface does not start
+    def apply_static6(ip, prefix, gw, dns, search = nil)
+      self.address6     = "#{ip}/#{prefix}"
+      self.gateway6     = gw
+      self.dns          = dns
+      self.search_order = search
```
carbonin: Should we have the same check here as we do in `#apply_static`?

isimluk: Nice catch, fixed. ✏️

isimluk:
> Should we remove some of the IPv6 specific fields in enable_dhcp?

I don't know for sure. Nothing should break if we don't.

From what I know today, it seems that dhcpv4 and static IPv6 could work side by side. I plan to dive deeper in DHCPv6 once the static stuff is merged.

It is not clear to me whether `BOOTPROTO='static'` applies to IPv6 or not. Needs some experiments.

carbonin:
Just kicking to make sure hakiri is fixed.

isimluk:
Rebased to get harakiri green.

spec/network_interface/network_interface_rh_spec.rb
```diff
@@ -100,6 +100,21 @@ def result(output, exit_status)
...
+  describe '#address6=' do
+    it 'sets the ipv6 address' do
+      address = 'fe80::1/64'
+      dhcp_interface.address6 = address
+      conf = dhcp_interface.interface_config
+      expect(conf['IPV6ADDR']).to eq(address)
+      expect(conf['BOOTPROTO']).to eq('static')
+      expect(conf['IPV6INIT']).to eq('yes')
+      expect(conf['DHCPV6C']).to eq('no')
+    end
```
carbonin: Just missing a space :smile:

isimluk:
Thanks!! I have brought in the lost new line.

isimluk:
Thanks!

PR #180, opened 2017-03-13
Add versioning to the nokogiri dependency

### LINKED ISSUE #179: Fix Hakiri issues
It's hard to review PRs because they all appear as failed.  Let's see if we can find why Hakiri fails, or just remove Hakiri. cc @carbonin

carbonin:
This should fix the Hakiri security warnings

Fixes #179

### DIFF
--- a/linux_admin.gemspec
+++ b/linux_admin.gemspec
@@ -38,7 +38,7 @@ registration, updates, etc.
   spec.add_dependency "awesome_spawn",        "~> 1.3"
   spec.add_dependency "inifile"
   spec.add_dependency "more_core_extensions", "~> 3.0"
-  spec.add_dependency "nokogiri"
+  spec.add_dependency "nokogiri",             ">= 1.7"
   spec.add_dependency "openscap"
   spec.add_dependency "net-ssh", "~> 3.2.0"
 end


isimluk:
I think this breaks my devel setup.

I have the following in my main Gemfile.dev.rb
```
['manageiq-ui-classic', 'manageiq-gems-pending', 'linux_admin', 'awesome_spawn'].each do |dep|
  dependencies.reject! { |d| d.name == dep }
  gem dep, :path => File.realpath('../' + dep, __dir__)
end
```

and it no longer resolves:

```
Bundler could not find compatible versions for gem "nokogiri":
  In Gemfile:
    handsoap (~> 0.2.5) was resolved to 0.2.5, which depends on
      nokogiri (>= 1.2.3)

    linux_admin was resolved to 0.19.0, which depends on
      nokogiri (>= 1.7)

    manageiq-gems-pending was resolved to 0.1.0, which depends on
      nokogiri (~> 1.6.8)

    rubywbem was resolved to 0.1.0, which depends on
      nokogiri (~> 1.5)
```

Perhaps we need to have the same harakiri check on `manageiq-gems-pending` repo to notify us about the need to update `nokogiri`?

carbonin:
Good catch, glad we saw this before we released.

@isimluk Want to open an issue on manageiq-gems-pending?

isimluk:
Looking into this a bit more. We should have used ">=1.6.8" instead of "1.7".

The 1.6.8 should contain all the fixes. And we wouldn't need bump manageiq-gems-pending, azure-armrest, handsoap, rubywbem and what not.

If I was famous Czech oracle - princess Lubossa, I would say I see a lot of headaches on the horizon given the `~>` operator and the reposplits.

isimluk:
https://github.com/ManageIQ/linux_admin/pull/183

PR #181, opened 2017-03-21
Network prefix-length vs netmask round 2

isimluk:
This is #176 reworked based on feedback.

@carbonin @jrafanie @Fryguy I tried to cherry pick the best from your feedback.


isimluk:
Thanks!

PR #185, opened 2017-03-23
Introduce enable_dhcp6 method

isimluk:
For enabling DHCPv6 client.

@miq-bot add_label enhancement

### DIFF
--- a/lib/linux_admin/network_interface/network_interface_rh.rb
+++ b/lib/linux_admin/network_interface/network_interface_rh.rb
@@ -95,13 +95,25 @@ module LinuxAdmin
     end
 
     # Set up the interface to use DHCP
-    # Removes any previously set static networking information
+    # Removes any previously set static IPv4 networking information
     def enable_dhcp
       @interface_config["BOOTPROTO"] = "dhcp"
       @interface_config.delete("IPADDR")
       @interface_config.delete("NETMASK")
       @interface_config.delete("GATEWAY")
       @interface_config.delete("PREFIX")
+      @interface_config.delete("DNS1")
+      @interface_config.delete("DNS2")
+      @interface_config.delete("DOMAIN")
+    end
+
+    # Set up the interface to use DHCPv6
+    # Removes any previously set static IPv6 networking information
+    def enable_dhcp6
+      @interface_config['IPV6INIT'] = 'yes'
+      @interface_config['DHCPV6C'] = 'yes'
+      @interface_config.delete('IPV6ADDR')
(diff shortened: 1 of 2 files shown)


carbonin:
Do we really want this to be separate?

I feel like enabling dhcp should be one concept and then dependent on what kind of dhcp server is on the user's network.

Do you think there is a use case for specifically enabling dhcpv4 and not dhcpv6 (or the other way around)?

isimluk:
I work on IPv6-only enablement. So, I guess having only single one stack is a thing. No strong opinion though. The downside is only having extra dhclient running. 

We may learn more once this gets some adoption, to a certain extent we are blind of how the set-up will look like.

People on the internet sometimes ask for ipv4 static and ipv6 dhcp settings or contrawise.

Look at https://github.com/ManageIQ/manageiq-gems-pending/pull/99, I think the dialog like this will be easy to comprehend for end-user.

carbonin:
> People on the internet sometimes ask for ipv4 static and ipv6 dhcp settings or contrawise.

I feel like the way this is now, we won't be able to support such a thing. For example, setting either protocol to dhcp removes the `DNS` and `DOMAIN` keys that may have been set when configuring static for the other protocol.

isimluk:
To a certain extent. The DNS should be managed by different set of methods in NetworkInterface. That is evidenced by calls of `resolv = LinuxAdmin::Dns.new; resolv.search_order = []; resolv.nameservers = []; resolv.save` in multiple places in `appliance_console`.

The thing underneath works like this:
 - dns entries in network-scripts gets appended to the `resolve.conf`
 - then dhcp, dhcpv6 appends to `resolve.conf`
 - `resolve.conf` can hold only three entries

It's lot of mess. I guess our code is only to help users to set-up simple networking scenarios when they first boot. I wish we never had this code on our own.

Should I merge the `enable_dhcp6` to enable_dhcp` then?

carbonin:
Let's do that then.

Probably a separate PR, but it sounds like we should back the `NetworkInterface#dns=` and `NetworkInterface#search_order=` methods with `LinuxAdmin::Dns` to make sure we have the "right" source of information.

From what you described, the network-scripts keys for these are just fronts for configuring `resolve.conf` properly so it feels like our abstraction should act the same way. Then maybe we remove the ability to mess with nameservers and search order entirely from network interfaces in the future.

For this PR I guess we can leave it the way it is. That way in the :sparkles: future :sparkles: when we implement :point_up: the DNS and DOMAIN keys will be removed from the network configuration if they were already there.

PR #186, opened 2017-04-19
LinuxAdmin.run/run! used to work, put it back via delegation

jrafanie:
This regressed here:
https://github.com/ManageIQ/linux_admin/commit/9e7d11322d793fe572a0cb94ffcded36bb4bfb94

Existing scripts that worked on 0.14.0 would be broken by 0.15.0+

### DIFF
--- a/lib/linux_admin.rb
+++ b/lib/linux_admin.rb
@@ -38,10 +38,13 @@ require 'forwardable'
 module LinuxAdmin
   class << self
     extend Forwardable
+    extend Gem::Deprecate
     attr_writer :logger
 
     def_delegators Common, :run
     def_delegators Common, :run!
+    deprecate :run,  "AwesomeSpawn.run",  2017, 6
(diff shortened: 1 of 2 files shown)


jrafanie:
@bdunne @carbonin can you review?

carbonin:
I know, I did that on purpose. What is this breaking?

ref: https://github.com/ManageIQ/manageiq/pull/6628

Fryguy:
Was it deprecated properly and released with a proper semver change?

Fryguy:
Since we are less than a 1.0 release, according to SemVer, we can *technically* break in a minor release, but the better way would have been to deprecate it until 1.0 where we break it.

Fryguy:
@jrafanie Can you please add a deprecation warning?

jrafanie:
> @jrafanie Can you please add a deprecation warning?

Done, rubygem's deprecate requires the month and year, which is useless... but it's a simple dsl baked into all modern rubies.

PR #188, opened 2017-07-14
Fix configure network fails after a successful DHCP configuration

ailisp:
**ISSUE**: After a successful DHCP configuration, if you configure network using static ipv6 it will fails. That fails is actually `ifdown` returns `1` indicate previous setting to network (DHCP) one is not successfully shutdown. However, actually it does if you run `ifconfig` at this time. This is why the network configuration becomes empty in appliance console status: `ifdown` fails so `NetworkInterfaceRH.save` returns `false` immediately without start up `eth0` using old configuration or new configuration. After `ifdown` returns `1` if we run another `ifdown` it will return `0` and subsequent `ifup` new configuration works.

**BZ**: https://bugzilla.redhat.com/show_bug.cgi?id=1464999

\cc @yrudman @carbonin 

@miq-bot add-label wip,bug

### DIFF
--- a/lib/linux_admin/network_interface/network_interface_rh.rb
+++ b/lib/linux_admin/network_interface/network_interface_rh.rb
@@ -161,7 +161,9 @@ module LinuxAdmin
     def save
       old_contents = File.read(@interface_file)
 
-      return false unless stop
+      stop_success = stop
+      stop_success = stop unless stop_success
+      return false unless stop_success
 
       File.write(@interface_file, @interface_config.delete_blanks.collect { |k, v| "#{k}=#{v}" }.join("\n"))
 
--- a/lib/linux_admin/network_interface/network_interface_rh.rb
+++ b/lib/linux_admin/network_interface/network_interface_rh.rb
@@ -162,6 +162,11 @@ module LinuxAdmin
       old_contents = File.read(@interface_file)
 
       stop_success = stop
+      # Stop twice because when configure both ipv4 and ipv6 as dhcp, ipv6 dhcp client will
+      # exit and leave a /var/run/dhclient6-eth0.pid file. Then stop (ifdown eth0) will try
+      # to kill this exited process so it returns 1. In the second call, this `.pid' file
+      # has been deleted and ifdown returns 0.
+      # See: https://bugzilla.redhat.com/show_bug.cgi?id=1472396
       stop_success = stop unless stop_success
       return false unless stop_success
 
--- a/spec/network_interface/network_interface_rh_spec.rb
+++ b/spec/network_interface/network_interface_rh_spec.rb
@@ -269,7 +269,7 @@ EOF
 
     it "returns false when the interface cannot be brought down" do
       expect(File).to receive(:read).with(iface_file)
-      expect(dhcp_interface).to receive(:stop).and_return(false)
+      expect(dhcp_interface).to receive(:stop).twice.and_return(false)
       expect(File).not_to receive(:write)
       expect(dhcp_interface.save).to be false
     end


ailisp:
What happens in `ifdown`:
screenshot_manageiq-20170706_2017-07-14_10 29 14
`ifdown` Success if run twice:
screenshot_manageiq-20170706_2017-07-14_10 46 41

yrudman:
@miq-bot add-label wip, bug, blocker

lib/linux_admin/network_interface.rb
```diff
@@ -148,7 +148,7 @@ def start
     #
     # @return [Boolean] whether the command succeeded or not
     def stop
-      Common.run(Common.cmd("ifdown"), :params => [@interface]).success?
+      Common.run(Common.cmd("ifconfig"), :params => [@interface, "down"]).success?
```
carbonin: From the `ifconfig` man-page:

```plain
NOTE
       This program is obsolete!  For replacement check ip addr and ip link.  For statistics use ip -s link.
```

Can we find the newer version of this process rather than using an obsolete tool?

carbonin: Additionally, while testing this I found that `ifconfig eth0 down` didn't actually bring the network interface down.

A subsequent call to `ifup eth0` failed because the interface was still active.

carbonin:
Specifically I did the following:

screenshot from 2017-07-17 15-06-29

lib/linux_admin/network_interface.rb
ailisp: Interesting. What is the `echo $?` after `ifconfig eth0 down` in your test?

carbonin: `0`, which is another reason it is so troubling ...

ailisp: I found `ip a` and `ifconfig` are living in two world. If you type `ifconfig` after a `ifconfig down`, `eth0` has disappeared. But in `ip a`, it's still here, same as you.
screenshot_manageiq-20170714_2017-07-17_15 56 50

ailisp: So if i use `ip` to shutdown/start an interface, and not use `ifup & ifdown` it should be fine. Let me test.

carbonin:
@ailisp even if you us `ip` directly it still looks like the behavior will not be the same. ref: https://serverfault.com/questions/603906/ip-link-set-not-assigning-ip-address-but-ifup-does

ailisp:
@carbonin Right. I also found `ip link set eth0 down` doesn't shutdown eth0 ...

ailisp:
As @carbonin discovered. The first `ifdown eth0` is because of it shut down processes listed in `.pid` file, including an dhcp client, which not exists. Thus it returns 1. In the second call, the pid file has been deleted, so it will return 0. It's safe to call it twice.

ailisp:
@carbonin tested after merge current master. Works fine.

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -162,7 +162,9 @@ def apply_static6(ip, prefix, gw, dns, search = nil)
     def save
       old_contents = File.read(@interface_file)
 
-      return false unless stop
+      stop_success = stop
+      stop_success = stop unless stop_success
+      return false unless stop_success
```
carbonin: So I opened a BZ that describes why we have to do this (https://bugzilla.redhat.com/show_bug.cgi?id=1472396)

Can you add a comment here with the BZ link explaining why we are trying twice?

ailisp: After configure both IPv4 and IPv6 as dhcp. The dhclient instance for IPv6 exits and leaves a file in /var/run/dhclient6-eth0.pid. Now if we run  `ifdown eth0` the command fails because ifdown-eth is failing to kill a process that is no-longer running. It will then delete that pid file. So in the second run `ifdown eth0` it will return 0 and success.

See also: 
https://bugzilla.redhat.com/show_bug.cgi?id=1472396

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -161,7 +161,14 @@ def apply_static6(ip, prefix, gw, dns, search = nil)
     def save
       old_contents = File.read(@interface_file)
 
-      return false unless stop
+      stop_success = stop
+      # Stop twice because when configure both ipv4 and ipv6 as dhcp, dhcp client for ipv6
+      # exit and leave a /var/run/dhclient6-eth0.pid file. Then stop (ifdown eth0) will try
+      # to kill this exited process so it returns -1. In the second call, this `.pid' file
```
ailisp: sorry for the typo. fixed

PR #190, opened 2017-09-26
Allow using NetworkInterface when the file doesn't exist yet

bdunne:
Also rescue failures for refreshing interfaces that don't exist
https://bugzilla.redhat.com/show_bug.cgi?id=1439345

### DIFF
--- a/lib/linux_admin/network_interface.rb
+++ b/lib/linux_admin/network_interface.rb
@@ -23,15 +23,12 @@ module LinuxAdmin
     # Creates an instance of the correct NetworkInterface subclass for the local distro
     def self.new(*args)
       self == LinuxAdmin::NetworkInterface ? dist_class.new(*args) : super
-    rescue MissingConfigurationFileError
-      NetworkInterfaceGeneric.new(*args)
     end
 
     # @return [String] the interface for networking operations
     attr_reader :interface
 
     # @param interface [String] Name of the network interface to manage
-    # @raise [NetworkInterfaceError] if network information cannot be retrieved
     def initialize(interface)
       @interface = interface
       reload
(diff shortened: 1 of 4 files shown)


lib/linux_admin/network_interface.rb
```diff
@@ -43,7 +41,11 @@ def initialize(interface)
     # @raise [NetworkInterfaceError] if network information cannot be retrieved
```
Fryguy: This needs to be updated.

Fryguy: And on line 32

spec/network_interface/network_interface_rh_spec.rb
```diff
@@ -39,6 +39,9 @@ def result(output, exit_status)
 
   subject(:dhcp_interface) do
     allow(File).to receive(:exist?).and_return(true)
+    stub_path = described_class.allocate.send(:path_to_interface_config_file, device_name)
+    allow(Pathname).to receive(:new).and_return(stub_path)
```
Fryguy: Isn't this *really* just allow_any_instance_of :wink:

Fryguy: I'm sorry this is just hacky as hell haha 😈

Fryguy: In review, I think path_to_interface_config_file should just be a class method.  Then you don't need hackery to go about calling a private method on a new instance, but trying to avoid side effects of initialize.

lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -178,7 +180,11 @@ def save
         return false
       end
 
-      true
+      reload
```
carbonin: :+1:

PR #191, opened 2017-09-26
Add platform attribute to Scap class

carbonin:
Newer versions of scap-security-guide (> 0.1.32) add new xccdf files which match our glob pattern, but do not contain the remediations for the rules we want to run.

If we edit one of these files rather than the one for our target platform, the rules will not be remediated properly.

Specifying the platform allows us to find the file we need.

https://bugzilla.redhat.com/show_bug.cgi?id=1493193

### DIFF
--- a/lib/linux_admin/scap.rb
+++ b/lib/linux_admin/scap.rb
@@ -5,6 +5,8 @@ module LinuxAdmin
     PROFILE_ID = "linux-admin-scap"
     SSG_XML_PATH = Pathname.new("/usr/share/xml/scap/ssg/content/")
 
+    attr_reader :platform
+
     def self.openscap_available?
       require 'openscap'
       true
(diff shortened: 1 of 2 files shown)


lib/linux_admin/scap.rb
```diff
@@ -44,28 +54,29 @@ def lockdown_profile(xccdf_file_path, profile_id)
...
+      Dir.glob(SSG_XML_PATH.join("ssg-#{platform}-#{type}.xml")).detect { |f| f =~ /ssg-\w+-#{type}.xml/ }
     end
 
     def tempdir
       @tempdir ||= Pathname.new(Dir.tmpdir)
     end
 
     def xccdf_file
-      @xccdf_file ||= self.class.xccdf_file
+      @xccdf_file ||= self.class.xccdf_file(platform)
```
Fryguy: If the platform changes via the accessor, then this cache should be cleared, right?  (same with oval_file)

Fryguy: Alternately, use attr_reader instead, then enforce that callers must set the platform on initialize.

carbonin: I'm for the `attr_reader` approach, I don't see someone reusing this object for different platforms so I think it makes sense to set it on initialization.

PR #192, opened 2017-09-28
Alter the scap module to use the DataStream file over the XCCDF one

carbonin:
Helps to address https://bugzilla.redhat.com/show_bug.cgi?id=1496279

Also, some clarification can be found in https://bugzilla.redhat.com/show_bug.cgi?id=1496279#c10

The gist of this is that the DataStream file is self contained, where the XCCDF file references other files to do its job. The bug was caused by us using the XCCDF file when a change was introduced that extracted the platform definitions into a file which was not present when we were doing remediations.

We decided it would be more maintainable to use the ds file moving forward to guard against other refactorings in scap-security-guide like this one.

### DIFF
--- a/lib/linux_admin/scap.rb
+++ b/lib/linux_admin/scap.rb
@@ -2,7 +2,7 @@ require 'nokogiri'
 
 module LinuxAdmin
   class Scap
-    PROFILE_ID = "linux-admin-scap"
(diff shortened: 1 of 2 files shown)


Fryguy:
Does this require a major version bump?

carbonin:
> Does this require a major version bump?

Yup, although http://semver.org/#spec-item-4 mentions that there are pretty much no rules for `0.y.z` versions if we don't want to go to `1.0.0`

Issue #187, opened 2017-04-19
Bump minimum nokogiri version ... again


carbonin:
[Security issue from Hakiri](https://hakiri.io/projects/c91a7f89e057f7/stacks/4d54d4ab0538e5/builds/765d9dba180cd5/warnings/7730fa56859218)

This needs 1.7.1 as the new minimum

carbonin:
@Fryguy Can we get the `hacktoberfest` label here for issues like this?

Fryguy:
This was an old issue, so the 1.7.1 number was old...this should now be `"~> 1.8", ">= 1.8.1"`

PR #194, opened 2017-10-13
Update minimum version of Nokogiri

DuskyPrawn:
Fixes security issue found by Hakiri.
Fixes https://github.com/ManageIQ/linux_admin/issues/187
https://hakiri.io/github/ManageIQ/linux_admin/master/d8db390d0a8c5d286c1d4c5d04901fd41bb65bfc/warnings?name=Command+Injection


carbonin:
To fix the hakiri warning I think we would have to add a restriction like the one in the [azure-armrest gem](https://github.com/ManageIQ/azure-armrest/blob/master/azure-armrest.gemspec#L25)

Right now, that is `~> 1.8, >= 1.8.1`

PR #195, opened 2017-10-16
Bumps net-ssh version from 3.2 to 4.2

petrblaho:
This is needed b/c net-ssh version 3.2 has issues with Ruby 2.4.

See https://github.com/net-ssh/net-ssh/issues/535

I discovered this while testing https://github.com/ManageIQ/manageiq-ui-classic/pull/2394 and https://github.com/ManageIQ/manageiq-providers-openstack/pull/118

### DIFF
--- a/lib/linux_admin/ssh.rb
+++ b/lib/linux_admin/ssh.rb
@@ -25,7 +25,7 @@ module LinuxAdmin
 
     def execute_commands(commands, agent_socket, stdin)
       result = nil
-      args = {:paranoid => false, :number_of_password_prompts => 0}
+      args = {:verify_host_key => false, :number_of_password_prompts => 0}
       if agent_socket
         args.merge!(:forward_agent              => true,
                     :agent_socket_factory       => -> { UNIXSocket.open(agent_socket) })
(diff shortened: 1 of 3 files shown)


carbonin:
We have this dependency specified in the ManageIQ repo, this will also need to be changed before this gem is released. (https://github.com/ManageIQ/manageiq/blob/master/Gemfile#L48)

carbonin:
Have you looked through the changes to the gem? It looks like we mostly are stubbing the calls in the specs.

After a quick glance I can see that a [parameter we are using](https://github.com/ManageIQ/linux_admin/blob/master/lib/linux_admin/ssh.rb#L28) has been deprecated:
> :paranoid => deprecated alias for :verify_host_key

Could you change that in this patch as well?

I didn't see any other changes that would break us, but if you could give it another set of eyes, that would be great.

carbonin:
@Fryguy @bdunne Do we want this is be less restrictive in some way?

I don't think we can still support `3.x` versions as the `:paranoid` option was removed.

I don't see any reason we can't go fully to 4.x ...

aufi:
@bdunne Hi, can we ask for releasing a new version of this gem to be able update it in https://github.com/ManageIQ/manageiq-gems-pending/blob/master/manageiq-gems-pending.gemspec in order to unblock https://github.com/ManageIQ/manageiq/pull/16216? Thanks!

cc @agrare

PR #200, opened 2018-03-13
Prioritize SubscriptionManager over Rhn, and other warning fixes

NickLaMuro:
Now the `LinuxAdmin::Rhn` has been deprecated in https://github.com/ManageIQ/linux_admin/pull/198 , this means that every instantiation of `LinuxAdmin::Rhn` will then add a deprecation warning to `STDOUT`.

Previously, this would also be triggered even if you used the public class methods on `LinuxAdmin::SubscriptionManager` (the deprecation's suggested replacement), because under the hood it determine which to use (`Rhn` or `SubscriptionManager`) by initializing each and calling their `registered?` methods.

Furthermore, these means if `Rhn` is registered on the system in addition to the preferred `SubscriptionManager`, it the `LinuxAdmin::SubscriptionManager.validate_credentials` class method would actually end up calling `LinuxAdmin::Rhn.validate_credentials` instead because of the preference.

The fix
-------

1. It changes the priority to favor `SubscriptionManager` over `Rhn`
2. Adds a new initial check that will check against the subclasss first before trying the other subclasses in the order of our choosing.
3. Fixes some deprecation warnings in specs
    - Removes printing the `"[DEPRECATION] 'LinuxAdmin::Rhn'..."` in the `RegistrationSystem` specs (skipped the other case of it for now)
    - Removes ruby warning for trying to access `@registration_type` without it being defined (can cause warnings in consumers of this gem)

Links
-----

* Related effort: https://github.com/ManageIQ/manageiq/pull/17153

### DIFF
--- a/spec/registration_system_spec.rb
+++ b/spec/registration_system_spec.rb
@@ -1,4 +1,11 @@
-describe LinuxAdmin::RegistrationSystem do
+shared_context "RegistrationSystem.registration_type stubbing", :registered_system_stubbing do
+  def stub_registered_to_system(*system)
+    allow_any_instance_of(LinuxAdmin::SubscriptionManager).to receive_messages(:registered? => system.include?(:sm))
+    allow_any_instance_of(LinuxAdmin::Rhn).to receive_messages(:registered? => system.include?(:rhn))
+  end
+end
+
+describe LinuxAdmin::RegistrationSystem, :registered_system_stubbing do
   context ".registration_type" do
     it "when registered Subscription Manager" do
       stub_registered_to_system(:sm)
@@ -82,9 +89,22 @@ describe LinuxAdmin::RegistrationSystem do
       expect { LinuxAdmin::RegistrationSystem.method_does_not_exist }.to raise_error(NoMethodError)
     end
   end
+end
 
-  def stub_registered_to_system(*system)
-    allow_any_instance_of(LinuxAdmin::SubscriptionManager).to receive_messages(:registered? => (system.include?(:sm)))
-    allow_any_instance_of(LinuxAdmin::Rhn).to receive_messages(:registered? => (system.include?(:rhn)))
+describe LinuxAdmin::SubscriptionManager, :registered_system_stubbing do
+  context ".registration_type" do
+    it "when registered both" do
+      stub_registered_to_system(:sm, :rhn)
+      expect(described_class.registration_type).to eq(LinuxAdmin::SubscriptionManager)
+    end
+  end
+end
+
+describe LinuxAdmin::Rhn, :registered_system_stubbing do
+  context ".registration_type" do
+    it "when registered both" do
+      stub_registered_to_system(:sm, :rhn)
+      expect(described_class.registration_type).to eq(LinuxAdmin::Rhn)
+    end
   end
 end
--- a/spec/registration_system_spec.rb
+++ b/spec/registration_system_spec.rb
@@ -5,9 +5,14 @@ describe LinuxAdmin::RegistrationSystem do
       expect(described_class.registration_type).to eq(LinuxAdmin::SubscriptionManager)
     end
 
-    it "when registered RHN" do
-      stub_registered_to_system(:sm, :rhn)
+    it "when registered RHN only" do
+      stub_registered_to_system(:rhn)
       expect(described_class.registration_type).to eq(LinuxAdmin::Rhn)
+    end
+
+    it "when registered both" do
+      stub_registered_to_system(:sm, :rhn)
+      expect(described_class.registration_type).to eq(LinuxAdmin::SubscriptionManager)
     end
 
     it "when unregistered" do
--- a/spec/registration_system_spec.rb
+++ b/spec/registration_system_spec.rb
@@ -1,4 +1,9 @@
 shared_context "RegistrationSystem.registration_type stubbing", :registered_system_stubbing do
+  # stub out `warn` messages from LinuxAdmin::Rhn
+  before do
+    allow_any_instance_of(LinuxAdmin::Rhn).to receive(:warn)
+  end
+
   def stub_registered_to_system(*system)
     allow_any_instance_of(LinuxAdmin::SubscriptionManager).to receive_messages(:registered? => system.include?(:sm))
     allow_any_instance_of(LinuxAdmin::Rhn).to receive_messages(:registered? => system.include?(:rhn))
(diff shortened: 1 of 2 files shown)


NickLaMuro:
Fixed Hakiri warnings in https://github.com/ManageIQ/linux_admin/pull/201

spec/registration_system_spec.rb
```diff
@@ -1,15 +1,32 @@
-describe LinuxAdmin::RegistrationSystem do
+shared_context "RegistrationSystem.registration_type stubbing", :registered_system_stubbing do
+  # stub out `warn` messages from LinuxAdmin::Rhn
+  before do
+    allow_any_instance_of(LinuxAdmin::Rhn).to receive(:warn)
+  end
+
+  def stub_registered_to_system(*system)
+    allow_any_instance_of(LinuxAdmin::SubscriptionManager).to receive_messages(:registered? => system.include?(:sm))
```
carbonin: Can we use something other than `allow_any_instance_of` here?

I think `allow(LinuxAdmin::SubscriptionManager).to receive(:new).and_return(double("SubscriptionManager", :registered? => system.include?(:sm))` would do the job as well, right?

carbonin: Ah, just saw this was only moved, thought this was new.

chrisarcand:
@NickLaMuro I think this is going about it the wrong way (and what I say here probably should have just been done with the original deprecation in #198, I acknowledge)

"SubscriptionManager is preferred, and uses Rhn under the hood as the only exception to the rule of 'Rhn is deprecated'" creates the need for these weird hacks. SubscriptionManager should be able to stand up on itself and Rhn utilize SubscriptionManager as a deprecated crutch for however long it's needed.

That is to say, any logic that we *need* for SubscriptionManager found within Rhn should be moved to SubscriptionManager. Any methods that are now missing in Rhn from the move should refer to SubscriptionManager and cause a deprecation warning as normal. 

Make sense? Is this a reason this approach wouldn't work better than this?

NickLaMuro:
> Make sense?

@chrisarcand To be honest, no, and I have reread your statement about 4-5 times to try to maybe get what you are saying (arguably might be a reading comprehension part on my end... but I digress).  That said, I will try and answer:

* * *

So just doing some quick git research, the `RegistrationSystem.registration_type_uncached` has been in place for 5 years now:

https://github.com/ManageIQ/linux_admin/commit/fe3b39d7#diff-09e060ae786b7916c51fde7ebb8ae488R24

And looks like it was changed to favor `Rhn` shortly after:

https://github.com/ManageIQ/linux_admin/commit/d79f2948


With that background known, what it seems like you are suggesting is that we basically re-write this `method_missing` business, which seems very... risky, for the goal of this being basically a patch fix.  If that is the end goal, I don't think I am the right person to be working on it.

I also feel like I am missing some context as to "why" `Rhn` is now deprecated, more specifically "now", but I can assume because `SubscriptionManager` has been the new tech for a while now and soon `Rhn` will be sunset (just my guess).  Clearly it was chosen to be the default at one point (see second commit linked above), but based on #198, it seems that doesn't make sense anymore, or even trying to keep the features of it around.

To me, it seems like `SubscriptionManager` can just completely replace `Rhn` AND the `RegistrationSystem` code entirely in `v2.0` of `linux_admin`, and we avoid this intermediate step (that I think you are suggesting) all together.

* * *

Again, that goes above and beyond the original intent of the PR, which was to remove deprecation warnings (and side-car'd in fixing some ruby warnings), because libraries should not emit warnings to consumers by just using them, only if they were used improperly.  The MIQ case linked in the description is a situation where it was being used correctly, but a warning was still omitted, so I was trying to fix the oversight, not re-architect the entire codebase.

spec/registration_system_spec.rb
NickLaMuro: Yeah... I wasn't a fan of it myself, but I figured changing it now was not a good idea.

NickLaMuro:
To all:

For what it is worth, I don't mind splitting up or omitting pieces of what I wrote here.  Fixing the deprecation warning when a user was using `SubscriptionManager` as suggested and still getting the warning was the main fix, and was "done" in the first commit.  Everything after that was just extra.

chrisarcand:
My comment was largely thinking in more general terms and now I see why in the code itself it's sort of lost. I'm not really understanding how registrations work and why that conditional has to be that way. @bdunne given your experience here do you see how one can decouple `Rhn` away entirely?

bdunne:
We could just kill it completely and release v2.0.0 ¯\_(ツ)_/¯

chrisarcand:
*Now* you're talkin'.

chrisarcand:
So the `SubscriptionManager` doesn't actually use `Rhn` at all, it really is just the check. From your reading description it didn't seem like it. We can do this change with @carbonin's feedback or just remove it as @bdunne said, whichever. Sorry for the noise.

NickLaMuro:
@chrisarcand well, here is the thing:  `SubscriptionManager` won't "use" `Rhn`'s class for pieces that are missing in `SubscrptionManager` when calling `.verify_credentials` (the class method, not the instance method, which is handled by `method_missing` in the `RegistrationSystem` base class), it will straight up give you an instance of `Rhn` back (assuming that `registered?` returns `true` for `Rhn`).  To me, that is a _**bug**_

Specifically if there happens to be a system where both` SubscriptionManager.new.registered?` and `Rhn.new.registered?` returns `true`, and we are forced into using the deprecated codebase.  In practice, this probably isn't the case, but in a test suite like ManageIQ, it becomes super cumbersome to stub out and work with both without knowing the internals of this lib to do so, which is something we shouldn't be passing on to other code bases.

* * *

So to summarize the first and second commits:

* The first favors `SubscriptionManager` when making our opinionated decision to choose the `RegistrationSystem` for you, with out really fixing the "both are valid on this system case"
* The second allows the user, despite our own opinions, to force a options when both are valid, but still fall back when to our order when their choice isn't valid.

I would argue that the first and second commits should both be merged to fix the bug, but arguably if we didn't want that much change for a patch release, I could see just merging the first commit.

* * *

That all said, if you peeps want to refactor this and "make it great again", in addition to or in favor of this PR, be my guest and I won't be offended in whichever route is pick.

But, I literally forked this yesterday, so if you think you can trick my statement of:

> "hey, you have a bug in your code, and you should feel bad"
>
> \- me

with a counter of:

> "no, **YOU** should feel bad because you didn't completely refactor it and make it the best it can be!"
> 
> \- @chrisarcand totally said this word for word

I ain't falling for that one.  Nope.  Nope.  Nuh-uh.  Peace out.

nope

jrafanie:
While this solves the deprecation warning, I'm in favor of deleting it instead if that's something we can do, therefore I prefer https://github.com/ManageIQ/linux_admin/pull/202

PR #202, opened 2018-03-14
Drop LinuxAdmin::RegistrationSystem::Rhn

bdunne:
An alternative to #200 
Requires a major version bump

cc @NickLaMuro @chrisarcand


NickLaMuro:
Well... @jrafanie is going to say yes by default because of all the code you deleted...

Anyway, I remain indifferent, and this works just as good as the other.  Though, I guess you were more "giving me a heads up" instead of asking for my opinion...

chrisarcand:
Let's do this one. I think deleted code is the clearest kind of code :trollface:

jrafanie:
> No code is faster than no code 👍

💯 

🔥 🚽 🗑 

Fix deprecation by deleting code.  👍

jrafanie:
@bdunne I just realized the commit message doesn't really say why we can remove rhn support.  Maybe saying "the server side rhn is now gone, so the client is unneeded" and linking to a bz or another link would be enough.

bdunne:
@jrafanie Commit message updated

lib/linux_admin/registration_system.rb
```diff
@@ -36,7 +35,7 @@ def self.registration_type_uncached
 
     def self.white_list_methods
       @white_list_methods ||= begin
-        all_methods = RegistrationSystem.instance_methods(false) + Rhn.instance_methods(false) + SubscriptionManager.instance_methods(false)
+        all_methods = RegistrationSystem.instance_methods(false) + SubscriptionManager.instance_methods(false)
```
NickLaMuro: Is there a reason you are keeping this this class in here at all?  The only value this provided was that there were two different `RegistrationSystem` classes in the code base prior where one could fallback to the other.  Now you just have a bunch of complexity for no reason.

Seems like you could ditch this class entirely, and move the `lib/linux_admin/registration_system/subscription_manager.rb` file up to the base folder.  Would allow us to get rid of all of this `method_missing` business.

jrafanie:
Wow, we let this get old quick...  This deprecation is loud and annoying now.  CAN HAZ DELETE CODE?

How do we do this?  @bdunne @NickLaMuro any opinions on keeping the code but removing the deprecation or going crazy like this PR and dropping it all together...  I'm all for deleting code for stuff we don't support anymore.  Thoughts?

NickLaMuro:
@jrafanie well, you are the only person with a "MAKE CHANGEZ PLZ!" review, I think that question should be redirected to you, huh?

Anyway, based on the commit message, I think this make sense with the addition of a major version bump.  👍

PR #204, opened 2018-09-27
Update Rubocop to add compatibility with new rules

saulotoledo:
At [version`0.53.0`](https://github.com/bbatsov/rubocop/blob/3c3e315b84df45845440a25cbd71a5b99214b21d/CHANGELOG.md#0530-2018-03-05), the `TrailingCommaInLiteral` rule at Rubocop has been split into`TrailingCommaInHashLiteral` and `TrailingCommaInArrayLiteral` and no longer exists. We need to choose which version we want to support in the project, since it is not possible to have both versions by having that options in your `rubocop.yml` file.

In order to use the new updated rules, this PR updates the current rubocop version to `0.53.0`.

Should be merged with:
- https://github.com/ManageIQ/guides/pull/303

### DIFF
--- a/linux_admin.gemspec
+++ b/linux_admin.gemspec
@@ -33,7 +33,7 @@ registration, updates, etc.
   spec.add_development_dependency "rake"
   spec.add_development_dependency "rspec",    "~> 3.0"
   spec.add_development_dependency "coveralls"
-  spec.add_development_dependency "rubocop"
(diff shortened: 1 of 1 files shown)


kbrock:
Is this still viable since 303 was closed?

djberg96:
I think we can close this since we're at rubocop version 0.69 now.

PR #205, opened 2019-05-10
Support nvme device partitions

carbonin:
Previously we would construct an incorrect path for nvme device partitions because they use a different naming scheme.

For example, the partition with id 1 on the device `/dev/sda` would have path `/dev/sda1` which is what we were implementing previously in `Partition#path`. nvme devices are different in that the first partition of device `/dev/nvme0n1` would not be `/dev/nvme0n11`, but `/dev/nvme0n1p1`.

This patch detects the device model from the parted output and uses the alternate path format if the device is nvme. Otherwise it falls back to the old path scheme whether we can detect a mode or not.

This should (finally) fix https://bugzilla.redhat.com/show_bug.cgi?id=1629853 when it gets released and added to the console.

### DIFF
--- a/lib/linux_admin/disk.rb
+++ b/lib/linux_admin/disk.rb
@@ -105,7 +105,7 @@ module LinuxAdmin
 
     def str_to_bytes(val, unit)
       case unit
-      when 'K' then
+      when 'K', 'k' then
         val.to_f * 1_024 # 1.kilobytes
       when 'M' then
         val.to_f * 1_048_576 # 1.megabyte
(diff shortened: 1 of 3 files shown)


carbonin:
I tried my best to fix the Hakiri failures in https://github.com/ManageIQ/linux_admin/pull/206, but one of the versions they want for nokogiri would conflict with gems-pending so I kept that out of this PR.

carbonin:
Also updated the ruby versions we use for testing in this PR because the bot wanted squiggly heredocs.

carbonin:
And had to remove the bundler constraint because I updated the ruby version ... all this for heredocs :weary:

PR #206, opened 2019-05-10
Modernize ruby stuff

carbonin:
Update bundler dependency, update travis.yml, update dependencies, etc.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,8 @@
 language: ruby
 rvm:
-- "2.1"
-- "2.2.6"
-- "2.3.3"
-- "2.4.0"
+- 2.4
+- 2.5
+- 2.6
 - ruby-head
 - jruby-head
 sudo: false
--- a/linux_admin.gemspec
+++ b/linux_admin.gemspec
@@ -37,7 +37,7 @@ registration, updates, etc.
   spec.add_dependency "awesome_spawn",        "~> 1.3"
   spec.add_dependency "inifile"
   spec.add_dependency "more_core_extensions", "~> 3.0"
-  spec.add_dependency "nokogiri",             ">= 1.8.5", "~> 1.8"
+  spec.add_dependency "nokogiri",             ">= 1.10.3", "~> 1.10"
   spec.add_dependency "openscap"
   spec.add_dependency "net-ssh", "~> 4.2.0"
 end
--- a/linux_admin.gemspec
+++ b/linux_admin.gemspec
@@ -37,7 +37,7 @@ registration, updates, etc.
   spec.add_dependency "awesome_spawn",        "~> 1.3"
   spec.add_dependency "inifile"
   spec.add_dependency "more_core_extensions", "~> 3.0"
-  spec.add_dependency "nokogiri",             ">= 1.10.3", "~> 1.10"
+  spec.add_dependency "nokogiri",             ">= 1.8.5", "!=1.10.0", "!=1.10.1", "!=1.10.2", "<2"
   spec.add_dependency "openscap"
   spec.add_dependency "net-ssh", "~> 4.2.0"
 end
--- a/linux_admin.gemspec
+++ b/linux_admin.gemspec
@@ -37,7 +37,7 @@ registration, updates, etc.
   spec.add_dependency "awesome_spawn",        "~> 1.3"
   spec.add_dependency "inifile"
   spec.add_dependency "more_core_extensions", "~> 3.0"
-  spec.add_dependency "nokogiri",             ">= 1.8.2", "~> 1.8"
+  spec.add_dependency "nokogiri",             ">= 1.8.5", "~> 1.8"
   spec.add_dependency "openscap"
   spec.add_dependency "net-ssh", "~> 4.2.0"
 end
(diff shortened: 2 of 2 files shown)


carbonin:
I have two commits here for nokogiri upgrades based on hakiri failures. That's because the later one upgrades it to a version which is incompatible with the one specified in `manageiq-gems-pending`. I'm not sure how keen we are to upgrade that one or maintain a released branch of this gem that doesn't require the newer version...

.travis.yml
```diff
@@ -1,9 +1,8 @@
...
-- "2.1"
-- "2.2.6"
-- "2.3.3"
-- "2.4.0"
+- 2.4
+- 2.5
+- 2.6
 - ruby-head
 - jruby-head
 sudo: false
```
jrafanie: I think this is the default now, but we can tackle that later

.travis.yml
```diff
@@ -1,9 +1,8 @@
 language: ruby
 rvm:
-- "2.1"
-- "2.2.6"
-- "2.3.3"
-- "2.4.0"
+- 2.4
+- 2.5
+- 2.6
```
jrafanie: 👍

linux_admin.gemspec
```diff
@@ -38,7 +37,7 @@ registration, updates, etc.
   spec.add_dependency "awesome_spawn",        "~> 1.3"
   spec.add_dependency "inifile"
   spec.add_dependency "more_core_extensions", "~> 3.0"
-  spec.add_dependency "nokogiri",             ">= 1.8.2", "~> 1.8"
+  spec.add_dependency "nokogiri",             ">= 1.10.3", "~> 1.10"
```
jrafanie: Warning, I have no looked at hakiri but @Fryguy schooled me in the ways of the ! operand with rubygems versions, in case you need to exclude specific old patch releases of a prior minor version while also wanting newer versions:

https://github.com/ManageIQ/manageiq-schema/commit/a45bd687499e1b4a76e57278aa6d01a6ff44f32f

carbonin: Okay, I'll take a look to see if I can do something similar to maintain compatibility with gems-pending and also fix hakiri here.

linux_admin.gemspec
```diff
@@ -38,7 +37,7 @@ registration, updates, etc.
   spec.add_dependency "awesome_spawn",        "~> 1.3"
   spec.add_dependency "inifile"
   spec.add_dependency "more_core_extensions", "~> 3.0"
-  spec.add_dependency "nokogiri",             ">= 1.8.2", "~> 1.8"
+  spec.add_dependency "nokogiri",             ">= 1.8.5", "!=1.10.0", "!=1.10.1", "!=1.10.2", "<2"
```
carbonin: @jrafanie @Fryguy this seems to solve the problem, but is there a more concise way of doing this? Alternatively, is there some documentation of the available operators? The only thing I can find is https://guides.rubygems.org/patterns/#pessimistic-version-constraint and that doesn't mention the `!=` at all.

.travis.yml
```diff
@@ -1,9 +1,8 @@
 language: ruby
 rvm:
-- "2.1"
-- "2.2.6"
-- "2.3.3"
-- "2.4.0"
+- 2.4
```
Fryguy: These should really be strings, otherwise they can cause issues on the Travis side as YAML parsing turns this into floats.

carbonin: They're not strings anywhere else ... https://github.com/ManageIQ/manageiq/blob/master/.travis.yml#L4-L5

Fryguy: Those are strings even though they aren't quoted...welcome to YAML parsing.

`key: 2` parses to Integer
`key: 2.4` parses to Float
`key: 2.4.0` parses to String

carbonin: :tired_face: Okay, I can update ... but for the record, travis is clearly fine with these as floats (the spec matrix ran fine).

carbonin: https://github.com/ManageIQ/linux_admin/pull/207

Fryguy: > travis is clearly fine with these as floats

That must be a more recent fix, because that always failed in the past...good to know :+1:

linux_admin.gemspec
jrafanie: Yay, I don't know.  It's smoke and mirrors... and some magic.

.travis.yml
jrafanie: lol, I didn't know this was a problem... I thought it was a consistency thing

jrafanie: Note, the blaster outer needs to do this correctly if we need these to be string literals.  It's currently putting bare values in the .travis.yml.

carbonin: Actually it looks like two digit versions would fail in the blaster outer https://github.com/ManageIQ/manageiq-release/blob/1be60fcafc4bf6ee499f54f853943424e7a7f327/scripts/pull_request_blaster_outer/change_travis_ruby_versions.rb#L4

carbonin: And to @Fryguy's point it looks like it would do the right thing and quote the two digit string:
```
irb(main):010:0> versions
=> ["2.3", "2.4", "2.5"]
irb(main):011:0> puts({"rvm" => versions}.to_yaml)
---
rvm:
- '2.3'
- '2.4'
- '2.5'
=> nil

```

jrafanie: yeah, I have to fix up the blaster outer... it's trying to do too much... it should really only be used for version bumps to a bunch of repos... I'll take a look

PR #208, opened 2019-07-05
Fix alignment and whitespace issues

bdunne:
Running the tests prior to the changes results in:

image

### DIFF
--- a/lib/linux_admin/fstab.rb
+++ b/lib/linux_admin/fstab.rb
@@ -56,18 +56,19 @@ module LinuxAdmin
   class FSTab
     include Singleton
 
-    attr_accessor :entries
-    attr_accessor :maximum_column_lengths
-
     def initialize
       refresh
     end
 
+    def entries
+      @entries ||= LinuxAdmin::FSTab::EntryCollection.new
+    end
+
     def write!
       content = ''
-      @entries.each do |entry|
+      entries.each do |entry|
         if entry.has_content?
-          content << entry.formatted_columns(@maximum_column_lengths) << "\n"
+          content << entry.formatted_columns(entries.maximum_column_lengths) << "\n"
         else
           content << "#{entry.comment}"
         end
@@ -84,20 +85,25 @@ module LinuxAdmin
     end
 
     def refresh
-      @entries  = LinuxAdmin::FSTab::EntryCollection.new
-      @maximum_column_lengths = Array.new(7, 0) # # of columns
+      @entries = nil
       read.each do |line|
         entry = FSTabEntry.from_line(line)
-        @entries << entry
+        entries << entry
       end
     end
 
     class EntryCollection < Array
+      attr_reader :maximum_column_lengths
+
+      def initialize
+        @maximum_column_lengths = Array.new(7, 0) # # of columns
+      end
+
       def <<(entry)
         lengths = entry.column_lengths
         lengths.each_index do |i|
-          @maximum_column_lengths[i] =
-            lengths[i] if lengths[i] > @maximum_column_lengths[i]
+          maximum_column_lengths[i] =
+            lengths[i] if lengths[i] > maximum_column_lengths[i]
         end
 
         super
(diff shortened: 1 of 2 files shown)


spec/fstab_spec.rb
```diff
@@ -63,4 +63,50 @@
       subject.instance.write!
     end
   end
+
+  describe "integration test" do
```
carbonin: This is still just testing `#write` though ... right?

spec/fstab_spec.rb
```diff
@@ -46,20 +46,84 @@
...
+                /dev/mapper/VG--MIQ-lv_os                     /  xfs defaults 0 0
+UUID=02bf07b5-2404-4779-b93c-d8eb7f2eedea                 /boot  xfs defaults 0 0
+              /dev/mapper/VG--MIQ-lv_home                 /home  xfs defaults 0 0
+               /dev/mapper/VG--MIQ-lv_tmp                  /tmp  xfs defaults 0 0
+               /dev/mapper/VG--MIQ-lv_var                  /var  xfs defaults 0 0
+           /dev/mapper/VG--MIQ-lv_var_log              /var/log  xfs defaults 0 0
+     /dev/mapper/VG--MIQ-lv_var_log_audit        /var/log/audit  xfs defaults 0 0
+               /dev/mapper/VG--MIQ-lv_log /var/www/miq/vmdb/log  xfs defaults 0 0
+              /dev/mapper/VG--MIQ-lv_swap                  swap swap defaults 0 0
+EOC
```
carbonin: Can you fix all of these to use `<<~` since we only support ruby >2.3 now?

carbonin: This should also fix a bunch of the rubocop issues I think.

carbonin:
Ah, also could you address the rubocop comments for the newly added code?

spec/fstab_spec.rb
bdunne: Not really, it is also testing the parsing, max column length logic, etc.

lib/linux_admin/fstab.rb
```diff
@@ -44,30 +44,31 @@ def columns
...
-        with_index { |col, i| col.to_s.rjust(max_lengths[i]) }.join(" ")
+        with_index { |col, i| col.to_s.rjust(max_lengths[i]) }.join(" ").rstrip
     end
   end
 
   class FSTab
     include Singleton
 
-    attr_accessor :entries
-    attr_accessor :maximum_column_lengths
```
carbonin: Maybe we should keep a method for this so we don't change API compatibility.

bdunne: Done

lib/linux_admin/fstab.rb
```diff
@@ -44,30 +44,35 @@ def columns
...
       refresh
     end
 
+    def entries
+      @entries ||= LinuxAdmin::FSTab::EntryCollection.new
+    end
+
+    def maximum_column_lengths
+      entries.maximum_column_lengths
+    end
```
jrafanie: Do we care about the setter methods that we're losing by implementing only the getter method?

carbonin: I don't even know what the setter would do in this case ... Nothing productive for sure.

bdunne: I don't think the setter should have ever been public api

carbonin:
@bdunne do you want to take care of the release?

simaishi:
@bdunne Travis is failing in manageiq-schema. Looks like this PR is the cause?

```
  1) DropMiqServerRhnMirror#up cleans up external files on an appliance
     Failure/Error: described_class.migrate(migration_direction)
       #<File (class)> received :write with unexpected arguments
         expected: ("/etc/fstab", "/dev/sda     / xfs defaults        0        0 \n")
              got: ("/etc/fstab", "/dev/sda     / xfs defaults 0 0\n")
```

https://travis-ci.org/ManageIQ/manageiq-schema/jobs/556038000

PR #209, opened 2019-09-27
Remove deprecated LinuxAdmin.run!

kbrock:
Remove deprecated `LinuxAdmin.run!` and `run`

Deprecation warning for tests are also removed.

### DIFF
--- a/lib/linux_admin.rb
+++ b/lib/linux_admin.rb
@@ -33,18 +33,10 @@ require 'linux_admin/ip_address'
 require 'linux_admin/dns'
 require 'linux_admin/network_interface'
 require 'linux_admin/chrony'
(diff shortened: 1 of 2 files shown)


kbrock:
As an alternative, I could revive #202 and create a new PR to delete `LinuxAdmin.run!`

let me know how to proceed

lib/linux_admin.rb
```diff
@@ -40,11 +40,6 @@ class << self
     extend Forwardable
```
bdunne: Do we still need `Forwardable` and `Gem::Deprectate`?

kbrock: works - thanks

PR #211, opened 2019-10-08
[WIP] Add disk size check

d-m-u:
'Choose the database disk: |1| '
    #    1) /dev/sr0: 0 MB
    #    2) /dev/vdb: 4768 MB
    #    3) Don't partition the disk

We shouldn't show /dev/sr0 in that list cause database creation on a SCSI CDROM is not a thing.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1693189

Er, yeah, it needs tests. But I wanted someone to shoot it down first I guess.

(Also I think we have bug in the size method this thing uses so this PR is fixing a symptom.)

@miq-bot add_label bug
@miq-bot assign @jrafanie

### DIFF
--- a/lib/linux_admin/disk.rb
+++ b/lib/linux_admin/disk.rb
@@ -12,7 +12,7 @@ module LinuxAdmin
       result = Common.run!(Common.cmd("lsblk"), :params => {:d => nil, :n => nil, :p => nil, :o => "NAME"})
       result.output.split.collect do |d|
         Disk.new :path => d
-      end
+      end.select { |disk| disk.size.nonzero? }
     end
 
     def initialize(args = {})


djberg96:
I'm curious why we have a custom `size` method, when we could modify the `lsblk` command like so:

`>lsblk -dnpb -o name,size`

On my system the size results were identical to the `fdisk -l` output. This would also have the advantage of not requiring sudo access.

djberg96:
@d-m-u What do you think about this?

```
  def self.local
    result = Common.run!(Common.cmd("lsblk"), :params => {:b => nil, :d => nil, :n => nil, :p => nil, :o => "NAME,SIZE"})
    result.output.split("\n").collect do |d|
      path, size = d.split
      self.new(:path => path, :size => size.to_i)
    end.select { |disk| disk.size.nonzero? }
  end

  def initialize(args = {})
    @path  = args[:path]
    @model = "unknown"
    @size  = args[:size]
  end
```

The modification to the -o option will add size to the output, which means we have to split on newlines instead of whitespace.

The `-b` option will return the size in bytes (same as `fdisk -l`)

The specs will require some minor updates but they're pretty superficial.

d-m-u:
I don't think we want to be using ```lsblk``` at all because it can't filter.

d-m-u:
Closing in favor of https://github.com/ManageIQ/linux_admin/pull/214 
Thanks @djberg96

PR #212, opened 2019-10-22
Add respond_to_missing to correspond with method_missing

djberg96:
Ruby best practices dictate that if you implement `method_missing` in a class, you should also implement a corresponding `respond_to_missing?` method.

So, this PR adds a `respond_to_missing?` private singleton method that checks for whitelisted methods. If found, it will return true, otherwise it will return false.

I've added a couple of specs as well.

Another reason to add this is that rspec will fail if run with strict partial validation in some of the core specs because dynamic methods aren't considered defined. You will see error messages like this:

```
19) MiqServer#enable_repos raises a notification for repos which fail to enable
      Failure/Error: expect(reg_system).to receive(:enable_repo).with("repo-1", anything).and_raise(err)
        LinuxAdmin::RegistrationSystem does not implement: enable_repo
      # ./spec/models/miq_server/update_management_spec.rb:217:in `block (3 levels) in <top (required)>'
```
Making this change eliminates those errors.

### DIFF
--- a/lib/linux_admin/registration_system.rb
+++ b/lib/linux_admin/registration_system.rb
@@ -24,6 +24,10 @@ module LinuxAdmin
 
     private
 
+    def self.respond_to_missing?(method_name, include_private = false)
+      white_list_methods.include?(method_name)
+    end
+
     def self.registration_type_uncached
       if SubscriptionManager.new.registered?
         SubscriptionManager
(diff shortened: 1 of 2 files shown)


djberg96:
@miq-bot add_label refactoring

djberg96:
@carbonin, @d-m-u look alright?

djberg96:
@miq-bot add_label enhancement

PR #214, opened 2019-10-30
Update Disk size and model collection

djberg96:
Followup to https://github.com/ManageIQ/linux_admin/pull/211 in case you want to stick with the `lsblk` command.

Here I add the size via the `-b` option (bytes), as well as include the model information. This should be more efficient, too, since it means it won't have to make a separate `fdisk -l` call to get the size.

I've added specs to confirm the behavior.

https://bugzilla.redhat.com/show_bug.cgi?id=1693189

### DIFF
--- a/lib/linux_admin/disk.rb
+++ b/lib/linux_admin/disk.rb
@@ -6,7 +6,7 @@ module LinuxAdmin
       [:id, :start_sector, :end_sector,
        :size, :partition_type, :fs_type]
 
-    attr_accessor :path, :model, :size
+    attr_accessor :path, :model
 
     def self.local
       result = Common.run!(Common.cmd("lsblk"), :params => {:b => nil, :d => nil, :n => nil, :p => nil, :o => "NAME,SIZE"})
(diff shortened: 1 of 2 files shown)


djberg96:
@miq-bot add_reviewer @carbonin

djberg96:
@d-m-u ok, updated. It will now collect the "type" and filter out anything that isn't a "disk".

As per our gitter conversation, I chose this over `-e 11` because I'm not sure device 11 is always going to be a cdrom type. Also, this approach filters out anything else that might accidentally creep in as opposed to just disks.

lib/linux_admin/disk.rb
```diff
@@ -8,16 +8,23 @@ class Disk
 
     attr_accessor :path, :model
 
+    # Collect local disk information via the lsblk command. Only disks with a
```
Fryguy: I don't think we should mention the implelementation of how we get the information in the docs.  LinuxAdmin is meant to be an abstraction, and we may choose a different tech for different stacks.

Fryguy:
cc @bdunne

d-m-u:
Um, ``` undefined local variable or method `string' for LinuxAdmin::Disk:Class (NameError)``` is what I get with this on a fresh appliance using the reproducing steps from the linked bz.

lib/linux_admin/disk.rb
```diff
@@ -8,16 +8,23 @@ class Disk
...
+    #
     def self.local
-      result = Common.run!(Common.cmd("lsblk"), :params => {:d => nil, :n => nil, :p => nil, :o => "NAME"})
-      result.output.split.collect do |d|
-        Disk.new :path => d
-      end
+      result = Common.run!(Common.cmd("lsblk"), :params => {:b => nil, :d => nil, :n => nil, :p => nil, :o => "NAME,SIZE,TYPE,MODEL"})
+      result.output.split("\n").collect do |string|
+        path, size, type, *model = string.split
+        if type.casecmp?('disk') && size.to_i > 0
```
d-m-u: Do you need to check if type is nil?

PR #223, opened 2020-11-16
drop ruby 2.4

d-m-u:
since we're not testing 2.4 in core, do we need it here?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: ruby
 rvm:
-- "2.4"
 - "2.5"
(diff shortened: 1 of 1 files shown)


Fryguy:
This is a standalone gem, so it can have different rules / support

d-m-u:
Support for it ended in March, yes?

Fryguy:
Yeah, but a) people still use it, and b) dropping it means having to cut a major version, which we probably won't do.  I'm not against dropping it, but we usually do so at those inflection points.

Issue #224, opened 2020-12-18
Linter warnings


djberg96:
Current list of linter warnings under lib:
```
Inspecting 39 files
...W.....W....W.W..W..WWW..W.......W.W.

Offenses:

lib/linux_admin/deb.rb:5:40: W: [Correctable] Lint/UnusedMethodArgument: Unused method argument - in_description. If it's necessary, use _ or _in_description as an argument name to indicate that it won't be used.
    def self.from_line(apt_cache_line, in_description=false)
                                       ^^^^^^^^^^^^^^
lib/linux_admin/fstab.rb:102:7: W: Lint/MissingSuper: Call super to initialize state of the parent class.
      def initialize ...
      ^^^^^^^^^^^^^^
lib/linux_admin/logical_volume.rb:43:5: W: Lint/MissingSuper: Call super to initialize state of the parent class.
    def initialize(args = {}) ...
    ^^^^^^^^^^^^^^^^^^^^^^^^^
lib/linux_admin/logical_volume.rb:57:5: W: [Correctable] Lint/UselessAccessModifier: Useless private access modifier.
    private
    ^^^^^^^
lib/linux_admin/logical_volume.rb:59:5: W: Lint/IneffectiveAccessModifier: private (on line 57) does not make singleton methods private. Use private_class_method or private inside a class << self block instead.
    def self.bytes_to_string(bytes)
    ^^^
lib/linux_admin/logical_volume.rb:71:5: W: [Correctable] Lint/UselessAccessModifier: Useless public access modifier.
    public
    ^^^^^^
lib/linux_admin/network_interface.rb:215:1: W: [Correctable] Lint/NonDeterministicRequireOrder: Sort files before requiring them.
Dir.glob(File.join(File.dirname(__FILE__), "network_interface", "*.rb")).each { |f| require f }
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
lib/linux_admin/null_logger.rb:5:5: W: Lint/MissingSuper: Call super to initialize state of the parent class.
    def initialize(*_args) ...
    ^^^^^^^^^^^^^^^^^^^^^^
lib/linux_admin/physical_volume.rb:22:5: W: Lint/MissingSuper: Call super to initialize state of the parent class.
    def initialize(args = {}) ...
    ^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #225, opened 2020-12-21
Some rubocop fixes

djberg96:
Some rubocop fixes. I just applied `rubocop -A` for this initial pass. Mostly it's unused variables. We can adjust as desired.

Mostly addresses https://github.com/ManageIQ/linux_admin/issues/224, but it does not yet fix the following:

```
lib/linux_admin/fstab.rb:102:7: W: Lint/MissingSuper: Call super to initialize state of the parent class.
      def initialize ...
      ^^^^^^^^^^^^^^
lib/linux_admin/logical_volume.rb:43:5: W: Lint/MissingSuper: Call super to initialize state of the parent class.
    def initialize(args = {}) ...
    ^^^^^^^^^^^^^^^^^^^^^^^^^
lib/linux_admin/null_logger.rb:5:5: W: Lint/MissingSuper: Call super to initialize state of the parent class.
    def initialize(*_args) ...
    ^^^^^^^^^^^^^^^^^^^^^^
lib/linux_admin/physical_volume.rb:22:5: W: Lint/MissingSuper: Call super to initialize state of the parent class.
    def initialize(args = {}) ...
```

Of those, the fstab one is trivial, but I'm not sure of the intent of the other 3.

### DIFF
--- a/lib/linux_admin/logical_volume.rb
+++ b/lib/linux_admin/logical_volume.rb
@@ -54,7 +54,6 @@ module LinuxAdmin
       self
     end
 
-    private
 
     def self.bytes_to_string(bytes)
       if bytes > 1_073_741_824 # 1.gigabytes
@@ -68,7 +67,6 @@ module LinuxAdmin
       end
     end
 
-    public
 
     def self.create(name, vg, value)
       self.scan # initialize local logical volumes
--- a/lib/linux_admin/logical_volume.rb
+++ b/lib/linux_admin/logical_volume.rb
@@ -54,7 +54,6 @@ module LinuxAdmin
       self
     end
 
-
     def self.bytes_to_string(bytes)
       if bytes > 1_073_741_824 # 1.gigabytes
         (bytes / 1_073_741_824).to_s + "G"
@@ -67,6 +66,7 @@ module LinuxAdmin
       end
     end
 
+    private_class_method :bytes_to_string
 
     def self.create(name, vg, value)
       self.scan # initialize local logical volumes
--- a/lib/linux_admin/service.rb
+++ b/lib/linux_admin/service.rb
@@ -8,7 +8,6 @@ module LinuxAdmin
     end
 
     class << self
-      private
       alias_method :orig_new, :new
     end
 
@@ -38,4 +37,4 @@ module LinuxAdmin
   end
 end
 
-Dir.glob(File.join(File.dirname(__FILE__), "service", "*.rb")).each { |f| require f }
+Dir.glob(File.join(File.dirname(__FILE__), "service", "*.rb")).sort.each { |f| require f }
--- a/lib/linux_admin/service.rb
+++ b/lib/linux_admin/service.rb
@@ -9,6 +9,7 @@ module LinuxAdmin
 
     class << self
       alias_method :orig_new, :new
+      private :orig_new
     end
 
     def self.new(*args)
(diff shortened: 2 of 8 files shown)


lib/linux_admin/logical_volume.rb
```diff
@@ -54,7 +54,6 @@ def extend_with(vg)
       self
     end
 
-    private
```
Fryguy, line 57: The intention here is that the bytes_to_string method was made private.  Can you add `private_class_method` to it?

lib/linux_admin/service.rb
```diff
@@ -8,7 +8,6 @@ def self.service_type(reload = false)
     end
 
     class << self
-      private
```
Fryguy, line 11: Not sure why this was removed?  I think this is intentional as the alias_method methods will be private, and that is no longer the case.

lib/linux_admin/volume.rb
```diff
@@ -1,6 +1,5 @@
 module LinuxAdmin
   class Volume
-    private
 
     def self.process_volume_display_line(line)
```
Fryguy, line 4: Same...Can you make this private_class_method (as well as elsewhere this type of change was made?)

Fryguy (changes requested):
private/protected had a meaning here (even if it wasn't actually working as expected), so it should be reflected in the class method.

lib/linux_admin/service.rb
djberg96: rubocop reported `Lint/UselessAccessModifier: Useless private access modifier`.

This script would seem to indicate that `private` doesn't work on aliases (?):

```ruby
class Foo
  class << self
    private
    alias_method :orig_new, :new
  end
end

p Foo.orig_new # works
```

djberg96: I think I've ended up resorting to just straight up `undef`-ing the `new` method when I wanted this behavior. Or used the `singleton` library (I think).

lib/linux_admin/volume.rb
djberg96: I did as many as I could without breaking the specs.

lib/linux_admin/service.rb
Fryguy: Interesting...I think private works with alias, but perhaps not alias_method ?

Either way, it is expected to be private so you could just do

```ruby
alias_method :orig_new, :new
private :orig_new
```

djberg96: Ok, updated.

lib/linux_admin/logical_volume.rb
djberg96: Done.

djberg96:
@Fryguy Look ok now?

lib/linux_admin/yum.rb
```diff
@@ -87,8 +87,6 @@ def self.repo_list(scope = "enabled")
       parse_repo_list_output(output)
     end
 
-    private
-
```
Fryguy, line 91: We should have private_class_method on all of the below methods to compensate for the intent.

djberg96: That breaks the current specs, so I couldn't tell if they were actually meant to be private.

djberg96:
I'll let someone else decide how to handle the public/private issue. Not critical, so closing.

PR #232, opened 2023-04-27
Fix Failing Build: update kwarg specs

kbrock:
Fixes failing build:

- `expect().with()` now distinguishes between kwargs and args. It requires `{}`
- Fixed all cops in one of the affected files.

PR #233, opened 2023-04-27
Use system path instead of hardcoding it.

kbrock:
depends upon:
- [x] https://github.com/ManageIQ/linux_admin/pull/232 (this is the first commit)

Introducing this so I can run tests on my mac.

alt:
- add `/opt/homebrew/bin` to `BIN_DIRS` which is in essence, the new `/usr/local/bin`

### DIFF
--- a/lib/linux_admin/common.rb
+++ b/lib/linux_admin/common.rb
@@ -4,7 +4,7 @@ module LinuxAdmin
   module Common
     include Logging
 
-    BIN_DIRS = %w(/bin /sbin /usr/bin /usr/sbin /usr/local/bin /usr/local/sbin)
+    BIN_DIRS = ENV["PATH"].split(File::PATH_SEPARATOR).freeze
 
     def self.cmd(name)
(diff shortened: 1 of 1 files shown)


kbrock:
update:
- rebased (should get rid of the cops - which were related to the commit from another PR)

kbrock:
@Fryguy @bdunne I feel the standard `PATH` mechanism from linux meets our needs.

This code started with absolute paths for various operating systems. So it makes sense that controlling the `PATH` ended up in the code. But now the path mechanism feels like reimplementing the os.

If there are concerns around ensuring the binary comes from the OS, much like systemd does, then I understand keeping this code in here, but then we probably wouldn't be leveraging `/usr/local/bin`. So that is why I vote removing this hardcoded path. Possibly we could remove the whole path mechanism, but that is more change/risky, and all I want is for this to work on my mac.

kbrock:
ok - thread said "Lets just use PATH"

agrare (approved):

PR #238, opened 2024-02-08
Fix AwesomeSpawn spec failures

kbrock:
AwesomeSpawn::CommandResult used to take 4 parameters but now takes 5
This was a change in the internal api for awesomespawn

In the future, we may want to converting over to using "awesome_spawn/spec_helper" and the stub_good_run stuff

### DIFF
--- a/spec/hosts_spec.rb
+++ b/spec/hosts_spec.rb
@@ -136,13 +136,13 @@ EOT
     end
 
     it "returns the hostname" do
-      result = AwesomeSpawn::CommandResult.new("", TEST_HOSTNAME, nil, 0)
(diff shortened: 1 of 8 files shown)


Fryguy:
Please also set the minimum version of awesome_spawn to 1.6, since that is the version that introduced this change.  That is, if someone has 1.5 and runs these specs, then it will fail in the opposite way.

agrare (approved):
:+1: good fix, similar to https://github.com/ManageIQ/manageiq-gems-pending/pull/568 and https://github.com/ManageIQ/manageiq-gems-pending/pull/569

kbrock:
I realize I was responsible for the change in the CommandResult parameters. (we should have added it to the end). Sorry

PR #244, opened 2024-04-16
Bump net-ssh for OpenSSL 3.0 support

agrare:
Also add support for `ed25519` type ssh keys

Related:
- https://github.com/ManageIQ/manageiq-appliance_console/issues/245
- https://github.com/ManageIQ/manageiq-ssh-util/pull/21

Required for: https://github.com/ManageIQ/manageiq-appliance_console/pull/248

### DIFF
--- a/linux_admin.gemspec
+++ b/linux_admin.gemspec
@@ -34,9 +34,11 @@ registration, updates, etc.
   spec.add_development_dependency "rubocop"
 
   spec.add_dependency "awesome_spawn",        "~> 1.6"
+  spec.add_dependency "bcrypt_pbkdf",         ">= 1.0", "< 2.0"
+  spec.add_dependency "ed25519",              ">= 1.2", "< 1.3"
   spec.add_dependency "inifile"
   spec.add_dependency "more_core_extensions", "~> 4.0"
-  spec.add_dependency "net-ssh", "~> 4.2.0"
+  spec.add_dependency "net-ssh",              "~> 7.2.3"
   spec.add_dependency "nokogiri",             ">= 1.8.5", "!=1.10.0", "!=1.10.1", "!=1.10.2", "<2"
   spec.add_dependency "openscap"
   spec.add_development_dependency "simplecov", ">= 0.21.2"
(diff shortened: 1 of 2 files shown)


agrare:
It looks like all versions of net-ssh which support openssl 3.0 require ruby >= 2.6
```
$ git tag --contains 8643e4e5b
v7.0.0
v7.0.0.beta1
v7.0.1
v7.1.0
v7.1.0.beta1
v7.1.0.beta2
v7.2.0
v7.2.0.beta1
v7.2.0.rc1
v7.2.1
v7.2.1.rc1
v7.2.2
v7.2.2.rc1
v7.2.3
```

linux_admin.gemspec
```diff
@@ -26,17 +26,19 @@ registration, updates, etc.
...
   spec.add_development_dependency "rspec",    "~> 3.0"
   spec.add_development_dependency "rubocop"
 
   spec.add_dependency "awesome_spawn",        "~> 1.6"
+  spec.add_dependency "bcrypt_pbkdf",         ">= 1.0", "< 2.0"
+  spec.add_dependency "ed25519",              ">= 1.2", "< 1.3"
   spec.add_dependency "inifile"
   spec.add_dependency "more_core_extensions", "~> 4.0"
-  spec.add_dependency "net-ssh", "~> 4.2.0"
+  spec.add_dependency "net-ssh",              "~> 7.2.3"
```
Fryguy, line 41: @bdunne Please review - do we know what the follow on effect of this is?
@agrare I'm wondering if we should release this as a major? (at a minimum a minor)

agrare: Since we need to drop a supported ruby versions yes this needs to be at least a minor bump

agrare: Breaking changes since 4.2.0:
`Breaking change: ed25519 now requires ed25519 gem instead of RbNaCl gem [#563]` - https://github.com/net-ssh/net-ssh/blob/master/CHANGES.txt#L127
`BREAKING: Drop support for Ruby 2.5` - https://github.com/net-ssh/net-ssh/blob/master/CHANGES.txt#L37

PR #249, opened 2024-08-01
Add LinuxAdmin::NetworkInterface.list

agrare:
Previously there wasn't a method for listing interfaces on the system, which was okay because we assumed an eth0 was available but as we move away from that and use the standard predictable network interface names it would be helpful to get a list.

This can be followed up with attributes indicating if an interface is a LOOPBACK type and if the interface is the dev registered for the default route.

Related https://github.com/ManageIQ/manageiq-appliance-build/pull/573

### DIFF
--- a/lib/linux_admin/network_interface.rb
+++ b/lib/linux_admin/network_interface.rb
@@ -20,6 +20,21 @@ module LinuxAdmin
       end
     end
 
+    def self.list
+      ip_link.pluck("ifname").map { |iface| new(iface) }
+    rescue AwesomeSpawn::CommandResultError => e
+      raise NetworkInterfaceError.new(e.message, e.result)
+    end
+
+    private_class_method def self.ip_link
+      require "json"
+
+      result = Common.run!(Common.cmd("ip"), :params => ["--json", "link"])
+      JSON.parse(result.output)
+    rescue AwesomeSpawn::CommandResultError, JSON::ParserError => e
+      raise NetworkInterfaceError.new(e.message, e.result)
+    end
+
     # Creates an instance of the correct NetworkInterface subclass for the local distro
     def self.new(*args)
       self == LinuxAdmin::NetworkInterface ? dist_class.new(*args) : super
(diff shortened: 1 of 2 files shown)


lib/linux_admin/network_interface.rb
```diff
@@ -20,6 +20,23 @@ def self.dist_class(clear_cache = false)
...
+        .map { |net| net["ifname"] }
+        .map { |iface| new(iface) }
+    rescue AwesomeSpawn::CommandResultError => e
+      raise NetworkInterfaceError.new(e.message, e.result)
+    end
+
+    private_class_method def self.ip_link
+      require "json"
+
+      ip_link = Common.run!(Common.cmd("ip"), :params => ["--json", "link"])
```
agrare, line 34: NOTE I think the `--json` output format would be helpful for e.g. `ip addr show` also which currently does some regex parsing of the output but I wanted to leave that for a standalone PR

Fryguy:
`ip` doesn't exist on a Mac - is that ok for this PR?

agrare:
> ip doesn't exist on a Mac - is that ok for this PR?

I think you're looking for the `OsxAdmin` repo :laughing: 

In all seriousness though, `LinuxAdmin::NetworkInterface#initialize` calls `ip addr show @interface`, and `ip route` already so I don't think this would work on a mac as is.

Fryguy:
code-wise LGTM - There are a lot of rubocop issues, and I wasn't sure if the new code was just following an existing pattern, or if this should be cleaned up.

agrare:
> There are a lot of rubocop issues, and I wasn't sure if the new code was just following an existing pattern

Yeah that is just how the specs are currently written so I wanted to keep it looking consistent.

lib/linux_admin/network_interface.rb
agrare: https://github.com/ManageIQ/linux_admin/pull/251

agrare:
@Fryguy I tackled the rubocop warnings here in https://github.com/ManageIQ/linux_admin/pull/252

Fryguy:
Merged #252 252, but now this is conflict

agrare:
Yes this added some ip command mocks so conflict was expected

agrare:
Rebased

Fryguy (approved):

PR #252, opened 2024-08-05
Fix rubocop warnings in NetworkInterface Spec

agrare:
Ref: https://github.com/ManageIQ/linux_admin/pull/249#issuecomment-2265444179


agrare:
NOTE we could make use of the AwesomeSpawn::SpecHelper stub_good_run!/stub_bad_run! here as well

Fryguy (approved):

PR #253, opened 2024-08-05
Parse config file on NetworkInterface#reload

agrare:
I haven't checked other callers of reload to see if this makes sense so I'd like your opinions @bdunne / @Fryguy but I was looking at https://github.com/ManageIQ/manageiq-appliance_console/blob/master/bin/appliance_console#L92-L93 and the `.reload` then `.parse_conf if respond_to?(:parse_conf)` looked like it could be improved by moving this logic into the NetworkInterfaceRh subclass.

### DIFF
--- a/lib/linux_admin/network_interface/network_interface_rh.rb
+++ b/lib/linux_admin/network_interface/network_interface_rh.rb
@@ -12,7 +12,15 @@ module LinuxAdmin
     def initialize(interface)
       @interface_file = self.class.path_to_interface_config_file(interface)
       super
+    end
+
+    # Gathers current network information for this interface
+    #
+    # @return [Boolean] true if network information was gathered successfully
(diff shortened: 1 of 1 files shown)


lib/linux_admin/network_interface/network_interface_rh.rb
```diff
@@ -12,6 +12,13 @@ class NetworkInterfaceRH < NetworkInterface
     def initialize(interface)
       @interface_file = self.class.path_to_interface_config_file(interface)
       super
```
agrare, line 14: NOTE the base class initialize calls `reload` so we don't need to super then parse_conf here.

Fryguy (approved):

Fryguy:
Yeah this makes more sense. 👍

PR #257, opened 2024-12-09
Fix ip_route on macOS

kbrock:
On typical systems, ip_route calls `ip route show default` On the mac, ip_route calls `ip route get while`

This fixes `ip_route` from failing and fixes an appliance_console bug

before
======

```
# LinuxAdmin::NetworkInterface.list
route: writing to routing socket: not in table (LinuxAdmin::NetworkInterfaceError)
 error was: route: writing to routing socket: not in table
 error was: route: writing to routing socket: not in table

	from /Users/kbrock/src/gems/linux_admin/lib/linux_admin/network_interface.rb:26:in `list'
	from (irb):2:in `<main>'
	from <internal:kernel>:187:in `loop'
	from railties (7.0.8.6) lib/rails/commands/console/console_command.rb:74:in `start'
	from railties (7.0.8.6) lib/rails/commands/console/console_command.rb:19:in `start'
```

after
=====

```
# LinuxAdmin::NetworkInterface.list.first
=>
#<LinuxAdmin::NetworkInterfaceDarwin:0x0000000163fd2d98
 @interface="lo0",
 @link_type="loopback",
 @network_conf={:address=>"127.0.0.1", :prefix=>8, :mac=>"00:00:00:00:00:00", :gateway4=>"192.168.1.1", :gateway6=>"fe80::%utun0"}>
```

ASIDE: I did start an update to ip_route, but it looks like it will not be a quick fix.


lib/linux_admin/homebrew.rb
```diff
@@ -0,0 +1,22 @@
...
+    def self.info(pkg)
+      out = Common.run!(homebrew_cmd, :params => ["list", :versions, pkg]).output
+      out.split("\n").each_with_object({}) do |line, pkg_hash|
+        name, ver = line.split(" ")[0..1] # only take the latest version
+        pkg_hash[name] = ver
+      end
+    end
+  end
+end
+
```
Fryguy, line 22: Look like you have a bonus line at the end here.

lib/linux_admin/network_interface.rb
```diff
@@ -15,6 +15,8 @@ def self.dist_class(clear_cache = false)
       @dist_class ||= begin
         if [Distros.rhel, Distros.fedora].include?(Distros.local)
           NetworkInterfaceRH
+        elsif Distros.local == Distros.darwin
+          NetworkInterfaceDarwin
```
Fryguy, line 19: I wonder if this is cleaner as:

```suggestion
        case Distros.local
        when Distros.rhel, Distros.fedora
          NetworkInterfaceRH
        when Distros.darwin
          NetworkInterfaceDarwin
```

Fryguy:
Overall looks good - can you add specs for ip_route method?

Fryguy:
My only other thought is, if we're going all the way to Darwin, should we try to avoid having to install iproute2mac, and use the built-in tools instead?

agrare:
I do think at a high level we need to answer "do we want linux_admin to work on mac?" (emphasis on linux here)

We can get some of the networking commands to work but there are some other command sets which will not (systemd vs old init.d service, rpm packages, etc...) and we could need something like supports here.

agrare:
I think the easy answer is "if it isn't difficult to get it to work we might as well" but how far do we want to go in order to get it to work?

Fryguy:
I'm ok with bringing in things bit by bit to support Mac.  Heck, I'd be ok with supporting WSL too at some point :)

Agreed on something like `supports` here would be good.  At a minimum we just raise NotImplementedError for certain things.

kbrock:
update:
- fixed case statement

kbrock:
@agrare I agree to the most part.

I don't want the appliance console to be able to administer the mac.
I do want the appliance console to start up and the menus to work.

I introduced these changes so I could get the basics of appliance console running to upgrade the highline gem.

kbrock:
update:
- fixed `split` to not pass `" "` as an argument but pass nothing

Fryguy (approved):

